PR #1, opened 2024-06-05
feat: terraform script added for Infrastructure(IAC)

sijandh35:
# Introducing Terraform for Infrastructure Management

This pull request (PR) introduces Terraform for managing our project's infrastructure. Terraform, developed by HashiCorp, is an Infrastructure as Code (IaC) tool that will help us automate and streamline our infrastructure provisioning and management. By integrating Terraform, we aim to achieve consistency, scalability, and efficiency in our infrastructure processes.

## Key Changes
- Added initial Terraform configuration files for [describe the specific components, e.g., "setting up AWS ECS, VPCs, and security groups"].
- Created Terraform modules to encapsulate reusable infrastructure components.
- Defined provider configurations and environment-specific variables.
- Updated documentation to include instructions on how to use Terraform for infrastructure management.

## Benefits
- **Consistency**: Ensures that our infrastructure is defined and provisioned in a consistent manner across all environments.
- **Automation**: Reduces the manual effort required to manage infrastructure, leading to faster deployments and fewer errors.
- **Scalability**: Facilitates easy scaling of infrastructure as our project grows.
- **Version Control**: Allows us to track and version our infrastructure changes using Git, providing a clear audit trail and the ability to revert to previous states if needed.

## Testing
- Verified Terraform configurations using `terraform plan` and `terraform apply` commands in a staging environment.
- Ensured that all resources are created and configured correctly according to our requirements.
- Conducted peer reviews(@nrjadkry, @mahesh-naxa) and incorporated feedback to improve the initial setup.

## Next Steps
- **Migrate Terraform configurations to OpenTofu, an open-source alternative for infrastructure management.**

## Notes
- Ensure you have the necessary permissions and API keys configured in your environment before running Terraform commands.

### DIFF
--- a/terraform/step4/REDIS.tf
+++ b/terraform/step4/REDIS.tf
@@ -1,0 +1,28 @@
+# Redis subnet group
+resource "aws_elasticache_subnet_group" "redis_subnet_grp" {
+  name        = "${var.project_name}-redis-subnet-grp"
+  description = "Subnet group for redis databases"
+  subnet_ids  = var.vpc_private_subnets[*]
+  tags = {
+    Name = "${var.project_name}-redis-subnet-grp"
+  }
+}
+
+
+# Redis-main
+resource "aws_elasticache_cluster" "redis-main" {
+  cluster_id           = "${var.project_name}-redis-main-cluster"
+  engine               = "redis"
+  node_type            = "cache.t2.small"
+  num_cache_nodes      = 1
+  parameter_group_name = "default.redis4.0"
+  engine_version       = "4.0.10"
+  port                 = 6379
+  subnet_group_name    = aws_elasticache_subnet_group.redis_subnet_grp.id
+  security_group_ids   = [var.redis_sec_grp]
+  apply_immediately    = true
+
+  tags = {
+    Name = "${var.project_name}-redis-main-cluster"
+  }
+}
--- a/.gitignore
+++ b/.gitignore
@@ -78,3 +78,57 @@ chart/charts
 
 #Docker
 DockerData/
+
+
+
+########################################## Terraform ##########################################
+# Terraform State files
+*.tfstate
+*.tfstate.*
+
+# Crash log files
+crash.log
+
+# Ignore override files as they are usually used to override resources locally
+override.tf
+override.tf.json
+*_override.tf
+*_override.tf.json
+
+# Ignore CLI configuration files
+.terraformrc
+terraform.rc
+
+# .terraform directory is used by Terraform to store module and provider caches
+.terraform/
+
+# Ignore any plan files
+*.tfplan
+
+# Ignore variables files that may contain sensitive information
+*.tfvars
+*.tfvars.json
+
+# Ignore local environment files
+.terraform.lock.hcl
+
+# Ignore backup files
+*.backup
+
+# Ignore editor specific files
+# Examples for Visual Studio Code, IntelliJ IDEA, etc.
+.vscode/
+.idea/
+
+# Ignore Mac OS specific files
+.DS_Store
+
+# Ignore Linux and Unix specific files
+*.swp
+*.swo
+
+#Ignore Pem File
+*.pem
+
+#Ignore Aws Cred
+*/.aws/credentials
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -34,22 +34,6 @@ repos:
         language: python
         types: [text]
 
-      - id: check-executables-have-shebangs
-        name: check that executables have shebangs
-        description: ensures that (non-binary) executables have a shebang.
-        entry: check-executables-have-shebangs
-        language: python
-        types: [text, executable]
-        stages: [commit, push, manual]
-
-      - id: check-shebang-scripts-are-executable
-        name: check that scripts with shebangs are executable
-        description: ensures that (non-binary) files with a shebang are executable.
-        entry: check-shebang-scripts-are-executable
-        language: python
-        types: [text]
-        stages: [commit, push, manual]
-
       - id: check-json
         name: check json
         description: checks json files for parseable syntax.
--- a/terraform/main.tf
+++ b/terraform/main.tf
@@ -1,0 +1,76 @@
+# Modules
+
+# VPC
+module "step1" {
+  source              = "./step1/"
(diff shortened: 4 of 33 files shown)


terraform/step4/REDIS.tf
```diff
@@ -0,0 +1,28 @@
...
+  description = "Subnet group for redis databases"
+  subnet_ids  = var.vpc_private_subnets[*]
+  tags = {
+    Name = "${var.project_name}-redis-subnet-grp"
+  }
+}
+
+
+# Redis-main
+resource "aws_elasticache_cluster" "redis-main" {
```
spwoodcock, line 13: What are we using Redis for in DroneTM?

spwoodcock (approved):
I have very little hands on experience with Terraform to be honest, but this looks sensible to me.

I do think it's a little insane how much config is required to set up an EC2 instance, S3, and DB 😂

Will this config change much over time / do you think it's maintainable?

spwoodcock:
Also, getting locked into an RDS managed database is a bad idea IMO.

If we deploy the same way as FMTM, we won't need it. 

I would like to deploy a managed database cluster ourselves via Kubernetes to be honest - it's not too tricky.

Could you give me a rough cost estimate for RDS for a year please so I can easily compare?

terraform/step4/REDIS.tf
sijandh35: @spwoodcock Initially i thought somewhere we would use celery for background task so i added configuration for Redis, For now i will comment this in the code and make a new PR.

sijandh35:
## Current Pricing Details

@spwoodcock  here is rough esitmate of our current setup from [AWS Calculator](https://calculator.aws/#/estimate?id=59d4bd805f4b5e71ee60a6d2a70b3ee563fc2001):

1. **Instance Cost:**
   - 1 instance(s) x 0.026 USD hourly x 730 hours in a month = 18.98 USD
   - Amazon RDS PostgreSQL instances cost (monthly): 18.98 USD

2. **Storage Cost:**
   - 100 GB per month x 0.131 USD = 13.10 USD
   - Storage pricing (monthly): 13.10 USD

**Total Estimated Monthly Cost:**

18.98 USD + 13.10 USD = 32.08 USD


*Note: This is a rough estimate. Additional charges such as data transfer costs may apply but are expected to be minimal.*

spwoodcock:
I would seriously doubt t3.micro would be good enough.

Maybe at the start if we just deploy the API without a proxy it's fine, but with more services it's unlikely
(2 core 1GB ram right?) We probably need more RAM

dakotabenjamin:
@spwoodcock , my thoughts on this is that its fine for IaC to be opinionated towards our needs or restrictions so long as sufficient documentation and images exist for others to build and deploy on their own infrastructure. I don't have an issue between running the database as a container vs RDS. Good "vendor-neutral" IaC should be engineered such that components are decoupled and either approach will work. Later down the line if we feel it beneficial to change (we need better backup management or IOPS controls), we can migrate the database over. 

A more pressing concern that I wished we had addressed earlier is that there seems to be some duplicated effort here with our existing [terraform modules](https://github.com/hotosm/?q=terraform&type=all&language=&sort=). Any chance you could look these over and see if there's some alignment with existing work?

Questions on design:
1. Why setup modules in steps? Doesn't terraform handle dependencies through 'requires'?
2. Does the EC2 need to be public and be issued an eip? Wouldn't the load balancer act as a public access point for that?
3. Can we manage sensitive info like passwords in Secrets Manager? 
4. On the topic of vendor locking, can we use ghcr.io rather than ECR?

spwoodcock:
Agree with your points!

As you suggested, we should all have a call to discuss the way forward 👍

sijandh35:
@dakotabenjamin @spwoodcock Thank you for the review lets discuss more on the call when everyone is free.cc @mahesh-naxa, @nischalstha9 .
1. **Why setup modules in steps? Doesn't terraform handle dependencies through 'requires'?**

We are using modules in Terraform to organize code by grouping similar services together. I have reviewed the Terraform modules you mentioned, where different repositories are created for each module. While that is a valid approach, it could introduce challenges with version management and state management across multiple repositories. Let’s discuss this further on our next call. If we plan to implement similar workflows across all HOT setups, organizing in that manner shouldn’t be an issue.

Additionally, we are planning to use Terragrunt or OpenTofu for deployment across different stages (dev, stage, prod). I hope this modular architecture will support that. What are your thoughts on this?


2. **Does the EC2 need to be public and be issued an eip? Wouldn't the load balancer act as a public access point for that?**

   Actually, EC2 is not needed here as we are using ECS for deployment. I am using EC2 just to connect to the database, which is on a private network, so I made EC2 also on a private network to connect to the database. However, this can be removed in the future as we can directly connect using ECS sessions. Currently, I am using an EC2 instance that falls within the free tier limit, which is t2.micro.

3. **Can we manage sensitive info like passwords in Secrets Manager?**

   Yes, that's a good idea. We will implement it on the pipeline.

4. **On the topic of vendor locking, can we use ghcr.io rather than ECR?**

   Okay, that won't be a problem.
   
And also, @spwoodcock, we have a complete setup that can be run on a single VM using different open-source services, which can also be migrated to managed vendor services and vice versa. Could you please review it? It’s in our repository.

spwoodcock:
Hey @sijandh35, I'm not sure we really discussed the Drone TM infrastructure as a team.

The only conversation I had about it was that NAXA would handle the deployments temporarily (i.e. a dev server), until HOT takes over in the future.

With this in mind, we would probably align our deployment with other tools at HOT, including the IaC.
We are of course very happy to collaborate on the IaC setup - your input would be very helpful for the team 🙏 

But for the short term, do you think it's best we leave setting this up until we need it & have time to review it all together?
The deployment could be done manually quite easily, if it's going to be shut down eventually.

mahesh-naxa:
Agreed @spwoodcock, i have been working on terraform modules on top of work that has been done by Yogesh, i believe we should setup a call so we can discuss and better align the work that we have been doing. Also i can use some help & direction as well. 
cc: @dakotabenjamin @sijandh35

nischalstha9:
Since @mahesh-naxa is planning to work on terraform modules and current infra is simplified on simple VM, we may need to revise infra code. Closed this PR since this branch has become stale.

PR #11, opened 2024-06-27
Dockerized Frontend with required changes to backend and docker stack

nischalstha9:
**What I have done?**

1. Dockerized frontend for development. (Production build still have issues with Media URLs)
2. Add URL prefix in backend FASTAPI
3. 404 Exception on backend returns html built with vite if built for live environment.
4. Read CORS origin from environment variable


src/backend/app/config.py
```diff
@@ -20,7 +21,7 @@ class Settings(BaseSettings):
     DEBUG: bool = False
     LOG_LEVEL: str = "INFO"
 
-    EXTRA_CORS_ORIGINS: Optional[Union[str, list[str]]] = []
+    EXTRA_CORS_ORIGINS: Optional[Union[str, list[str]]] = os.environ.get("EXTRA_CORS_ORIGINS","").split(",")
```
nrjadkry: We don't need to accept the environment like this. It does look at the EXTRA_CORS_ORIGINS by itself.

nischalstha9: Thanks, Didn't knew it was possible with `pydantic-settings`. I have updated the code!

nrjadkry (approved):

PR #17, opened 2024-07-01
Refactor: read_project endpoint to optimize database queries

Pradip-p:
## What type of PR is this?

- [ ] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation
- [x] 🧑‍💻 Refactor
- [ ] ✅ Test
- [ ] 🤖 Build or CI

## Describe this PR

Refactor `read_project` endpoint to optimize database queries and improve error handling:

- Updated SQL queries to efficiently retrieve project details and associated tasks.
- Enhanced error handling for cases where the project is not found.
- Replaced legacy async conversion methods with direct usage of Pydantic models for task conversion.

## Screenshots
image

No other approaches were considered for this refactor.

### How to test this change?
1. Set up a local environment with the required database setup.
2. Execute the `read_project` endpoint with valid and invalid project IDs.
3. Verify that projects are retrieved correctly with associated tasks.
4. Ensure appropriate error responses are returned when a project is not found.

### DIFF
--- a/src/backend/app/projects/project_crud.py
+++ b/src/backend/app/projects/project_crud.py
@@ -68,7 +68,6 @@ async def get_project_by_id(
     db: Database = Depends(database.encode_db), project_id: Optional[int] = None
 ):
     """Get a single project &  all associated tasks by ID."""
-    # check the project in Database
     raw_sql = """
     SELECT
         projects.id,
@@ -88,17 +87,9 @@ async def get_project_by_id(
     project_record = await db.fetch_one(raw_sql, {"project_id": project_id})
     query = """ SELECT * from tasks WHERE project_id = :project_id;"""
     task_records = await db.fetch_all(query, {"project_id": project_id})
-   
-    return  {
-        "id": project_record["id"],
-        "name": project_record["name"],
-        "short_description": project_record["short_description"],
-        "description": project_record["description"],
-        "per_task_instructions": project_record["per_task_instructions"],
-        "outline": project_record["outline"],
-        "tasks": task_records,
-        "task_count": len(task_records)
-    }
+    project_record.tasks = task_records
+    project_record.task_count = len(task_records)
+    return project_record
 
 
 async def get_projects(
--- a/src/backend/app/projects/project_crud.py
+++ b/src/backend/app/projects/project_crud.py
@@ -72,9 +72,12 @@ async def get_project_by_id(
     raw_sql = """
     SELECT
         projects.id,
+        projects.name,
         projects.short_description,
         projects.description,
+        projects.per_task_instructions,
         projects.author_id,
+        projects.outline,
         users.name AS author_name
     FROM projects
     JOIN users ON projects.author_id = users.id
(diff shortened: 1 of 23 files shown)


src/backend/app/projects/project_crud.py
```diff
@@ -54,15 +55,58 @@ async def create_project_with_project_info(
...
+    tasks =   await convert_to_app_tasks(all_tasks)
+    # Count the total tasks
+    task_count = len(tasks)
+    return {
+        "project": project,
+        "task_count": task_count,
+        "tasks": tasks,
+    }
+    
+async def convert_to_app_tasks(
```
spwoodcock: This code is not necessary with Pydantic. The logic should be directly in the models instead of passed through a function afterwards 👍

FMTM actually needs this refactored out at some time too

spwoodcock (changes requested):

src/backend/app/projects/project_crud.py
Pradip-p: I have made changes to the code so that the logic is now directly in the Pydantic models. Additionally, I have removed the f-string from the SQL query. Let me know if there is anything else needed.

src/backend/app/projects/project_crud.py
```diff
@@ -13,56 +12,115 @@
...
+    db_tasks: List[db_models.DbTask],
+) -> List[project_schemas.TaskOut]:
+    """Legacy function to convert db models --> Pydantic.
+
+    TODO refactor to use Pydantic model methods instead.
+    """
+    if db_tasks and len(db_tasks) > 0:
+
+        async def convert_task(task):
```
spwoodcock: So this kind of thing is also not necessary.

The way FastAPI and Pydantic work together is like this:
- Define your schemas in Pydantic
- Combine together the schemas however you need to make a model that is returned from an endpoint
- Specify the `response_model` on the endpoint.
- FastAPI / Pydantic handle the validation and serialisation for you.

So say you had a task model:

```python
TaskOut(BaseModel):
    something: str
    something2: int
```

then you can use this in your Project model:

```python
ProjectOut(BaseModel):
    id: int
    tasks: list[TaskOut]
```

Then you can use that in your endpoint:

```python
@app.get("/projects", response_model=list[ProjectOut])
...
    # logic here that queries the db and returns the db data, including associated tasks
    return db_projects
```

- Then when the endpoint is called, the projects are returned from the db.
- FastAPI will use Pydantic to serialise the db data into the format specified in the model (i.e. omits fields it doesn't need and includes the fields it does).
- FMTM has some examples of this that could help 👍

spwoodcock (changes requested):

spwoodcock:
Sorry to be a pain 😅 The changes are good, but there is just a minor thing!

See my comment on best practice FastAPI / Pydantic usage 👍

src/backend/app/projects/project_crud.py
Pradip-p:
```
@router.get("/{project_id}", tags=["Projects"], response_model=project_schemas.ProjectOut)
async def read_project(
    project_id: int,
    db: Database = Depends(database.encode_db),
):
    """Get a specific project and all associated tasks by ID."""
    project = await project_crud.get_project_by_id(db, project_id)
    if project is None:
        raise HTTPException(status_code=404, detail="Project not found")
    return project
```
```
async def get_project_by_id(
    db: Database = Depends(database.encode_db), project_id: Optional[int] = None
):
    """Get a single project &  all associated tasks by ID."""
    # check the project in Database
    raw_sql = """
    SELECT
        projects.id,
        projects.name,
        projects.short_description,
        projects.description,
        projects.per_task_instructions,
        projects.author_id,
        projects.outline,
        users.name AS author_name
    FROM projects
    JOIN users ON projects.author_id = users.id
    WHERE projects.id = :project_id
    LIMIT 1;
    """

    project_record = await db.fetch_one(raw_sql, {"project_id": project_id})
    # Fetch associated tasks
    query = """ SELECT * from tasks WHERE project_id = :project_id;"""
    task_records = await db.fetch_all(query, {"project_id": project_id})
   
    return  {
        "id": project_record["id"],
        "name": project_record["name"],
        "short_description": project_record["short_description"],
        "description": project_record["description"],
        "per_task_instructions": project_record["per_task_instructions"],
        "outline": project_record["outline"],
        "tasks": task_records,
        "task_count": len(task_records)
    }
```
```
class TaskOut(BaseModel):
    """Base project model."""

    id: int
    project_task_index: int
    project_task_name: Optional[str] = None
    outline: Any = Field(exclude=True)

    @computed_field
    @property
    def outline_geojson(self) -> Optional[Feature]:
        """Compute the geojson outline from WKBElement outline."""
        if not self.outline:
            return None
        wkb_data = bytes.fromhex(self.outline)
        geom = wkb.loads(wkb_data)
        bbox = geom.bounds  # Calculate bounding box
        return str_to_geojson(self.outline, {"id": self.id, "bbox": bbox}, self.id)

```
```
class ProjectOut(BaseModel):
    """Base project model."""

    id: int
    name: str
    short_description: str
    description: str
    per_task_instructions: Optional[str] = None
    outline: Any = Field(exclude=True)
    tasks: list[TaskOut] = None
    task_count: int = None

    @computed_field
    @property
    def outline_geojson(self) -> Optional[Feature]:
        """Compute the geojson outline from WKBElement outline."""
        if not self.outline:
            return None
        wkb_data = bytes.fromhex(self.outline)
        geom = wkb.loads(wkb_data)
        bbox = geom.bounds  # Calculate bounding box
        return str_to_geojson(self.outline, {"id": self.id, "bbox": bbox}, self.id)
```
@spwoodcock  I've been working on implementing project details along with their associated tasks and counting the total tasks. As you can see above, I'm using raw SQL to fetch projects and their associated tasks. While this approach aligns with typical FastAPI and Pydantic usage patterns, I feel somewhat uncomfortable with the structure of the get_project_by_id function's return statement:
 ```
{
    "id": project_record["id"],
    "name": project_record["name"],
    "short_description": project_record["short_description"],
    "description": project_record["description"],
    "per_task_instructions": project_record["per_task_instructions"],
    "outline": project_record["outline"],
    "tasks": task_records,
    "task_count": len(task_records)
}
```
I'm unsure if this is the best approach. Could you provide guidance or suggest an alternative method to transform raw data into a format that validates against a Pydantic schema?

Comment (spwoodcock) [reply src/backend/app/projects/project_crud.py]: This is lots better - great job!

You are right the current returned dict is prone to breaking if fields change.
You could simply:

```python
return await db.fetch_all(query, {"project_id": project_id}) 
```

This will pass the db object through the pydantic schema, filtering out fields not required and running the validators / serializers to JSON for the endpoint response 😄

FastAPI / Pydantic are neat

PR #20, opened 2024-07-02
Update Database Schemas

nrjadkry:
## Description
This pull request updates the database schemas in accordance with the design specifications outlined here [DTM db design](https://github.com/hotosm/datm-research/tree/main/db_design). by @spwoodcock 

### Changes:

- **Schema Modifications:**
  - Updated tables and relationships as per the new design.
  - Revised column definitions and data types to match the latest specifications.
  - Implemented new indexing strategies for improved query performance.
  
- **New Tables:**
  -  Following new tables are added
      1. Drone
      2. DroneFlight 
      3. TaskEvent

- **Data Migration:**
    I have used alembic migration for this. We can remove this in the future if we felt like we dont need it. I have created an initial migration script as well.


src/backend/app/db/db_models.py
```diff
@@ -62,95 +60,50 @@ class DbOrganisation(Base):
...
+        str, Column(UUID(as_uuid=True), ForeignKey("projects.id"), nullable=False)
     )
     project_task_index = cast(int, Column(Integer))
-    project_task_name = cast(str, Column(String))
     outline = cast(WKBElement, Column(Geometry("POLYGON", srid=4326)))
     task_status = cast(TaskStatus, Column(Enum(TaskStatus), default=TaskStatus.READY))
     locked_by = cast(
-        int,
-        Column(BigInteger, ForeignKey("users.id", name="fk_users_locked"), index=True),
+        str,
```
spwoodcock: I would remove these fields `locked_by`, `mapped_by`, `validated_by` as they are unnecessarily duplicated data across tables.

We have this data in task_events and just need a query to access them 👍 

For example:

```sql
SELECT user_id
FROM public.task_events
WHERE project_id=$1 AND task_id=$2 AND state='LOCKED_FOR_MAPPING'
ORDER BY event_id DESC
LIMIT 1;
```

src/backend/app/migrations/versions/d2b9d45d3ede_initial_migration.py
```diff
@@ -174,93 +202,122 @@ def upgrade() -> None:
...
+        sa.Column("event_id", sa.UUID(), nullable=False),
+        sa.Column("project_id", sa.UUID(), nullable=False),
+        sa.Column("task_id", sa.UUID(), nullable=False),
+        sa.Column("user_id", sa.String(length=100), nullable=False),
+        sa.Column("detail", sa.String(), nullable=True),
+        sa.Column(
+            "state",
             sa.Enum(
-                "RELEASED_FOR_MAPPING",
+                "READY",
```
spwoodcock, line 261: I think these existing enums should be replaced with:

```python
class State(int, Enum):
    """The state of an object (task or entity).

    The state can be:
    - ``unlocked to map``
    - ``locked for mapping``
    - ``unlocked to validate``
    - ``locked for validation``
    - ``unlocked done``
    """
    UNLOCKED_TO_MAP = 0
    LOCKED_FOR_MAPPING = 1
    UNLOCKED_TO_VALIDATE = 2
    LOCKED_FOR_VALIDATION = 3
    UNLOCKED_DONE = 4
```

```python
class EventType(str, Enum):
    """Events that can be used via the API to update a state

    Specify the event type for ``POST`` to:
    ``/project/{pid}/event``
    or for Entities endpoints too.

    Possible values are:

    - ``map`` -- Set to *locked for mapping*, i.e. mapping in progress.
    - ``finish`` -- Set to *unlocked to validate*, i.e. is mapped.
    - ``validate`` -- Request recent task ready to be validate.
    - ``good`` -- Set the state to *unlocked done*.
    - ``bad`` -- Set the state *unlocked to map* again, to be mapped once again.
    - ``split`` -- Set the state *unlocked done* then generate additional subdivided task areas.
    - ``assign`` -- For a requester user to assign a task to another user. Set the state *locked for mapping* passing in the required user id.
    - ``comment`` -- Keep the state the same, but simply add a comment.

    Note that ``task_id`` must be specified in the endpoint too.
    """
    MAP = "map"
    FINISH = "finish"
    VALIDATE= "validate"
    GOOD = "good"
    BAD = "bad"
    SPLIT = "split"
    ASSIGN= "assign"
    COMMENT = "comment"
```

src/backend/app/db/db_models.py
```diff
@@ -116,6 +116,7 @@ class DbProject(Base):
     organisation = relationship(DbOrganisation, backref="projects")
 
     # flight params
+    overlap = cast(float, Column(Float, nullable=True))  # in cm_px
     gsd = cast(float, Column(Float, nullable=True))  # in cm_px
```
spwoodcock: try to be precise in the namings for clarity (including units) - I used the names in did in the suggested schema for this purpose:
- ground_sample_distance_cm_px
- overlap_percent
- etc

spwoodcock: Someone looking at the database in the future will probably guess the units for `gsd` (if they know what it stands for).

By it might be harder to know in the Drones table that `weight` could be kg, american lbs, etc...

PR #65, opened 2024-07-10
Send email notifications to Project Creator and Drone Pilots

nrjadkry:
## Description
The requirement is to send email notification to the project creators whenever a drone pilot requests to map a task and after the project creator accepts it,  similar notification should be send to the drone pilot that his request has been accepted.

### Package Used
I have used [aiosmtplib](https://aiosmtplib.readthedocs.io/en/latest/) package to send the email to the users

### Tasks
- [x] SMTP Configurations
- [x] Utils functions to send the notifications
- [x] Email notification to the project creator
- [ ] Email notification to the drone operators


spwoodcock:
`python-emails` isn't bad, but it hadn't been updated in a while.

With FastAPI ideally we want to use an async mailer to prevent thread blocking. 

`aoismtplib` is good for this use case.
Example implementation using jinja templates:
 https://gitlab.com/mastweb/backend/-/blob/main/app/api/user_public.py?ref_type=heads#L133

nrjadkry:
@spwoodcock We could use `aoismtplib` for async mailer. I was thinking of adding the email task into background task. Using async mailer sounds good

nrjadkry:
@spwoodcock I have updated this to use `aiosmtp`. Will you take a look into it.

spwoodcock (approved):
Good stuff 👍

PR #71, opened 2024-07-13
Feat project creation

Pradip-p:
### Description

This PR introduces several enhancements and fixes to the project schema:

1. **New Features:**
   - **Deadline Column:** Added a new `deadline` column to the projects table to track submission deadlines.
   - **No Fly Zones Support:** Enhanced the `no_fly_zones` input to support `Polygon`, `MultiPolygon`, and other geometry types, ensuring flexible and accurate representation of restricted areas.
   **Additional Schema Fields**: Introduced the following new fields to the project schema:
        -  gsd_cm_px
        - is_terrain_follow
        - outline_no_fly_zones
        - output_orthophoto_url
        - output_pointcloud_url
        - output_raw_url
        - deadline

2. **Utility Functions:**
   - Added a `multipolygon_to_polygon` utility function to handle and convert multipolygon geometries within `no_fly_zones`.


spwoodcock:
Always good db practice to at `_at` for date columns, do `deadline_at` 😄

Pradip-p:
> Always good db practice to at `_at` for date columns, so `deadline_at` 😄
@spwoodcock  Thanks for the feedback! Just updated the column names with '_at' for better clarity. 😄

spwoodcock (approved):
Looks good 👍

Minor comment: to save listing this huge list of variables to inject, could the ones from `project_metadata` be inserted simply using `**project_metadata`?

They will be used as keyword args, so this will work where the variable names match.

It would also make the code more resilient to change / breakage if variables change in future.

src/backend/app/projects/project_crud.py
```diff
@@ -31,35 +30,44 @@ async def create_project_with_project_info(
...
-            "dem_url": project_metadata.dem_url,
-        },
-    )
+    try:
+        project_id = await db.execute(
+            query,
+            values={
+                "id": _id,
+                "author_id": author_id,
+                "name": project_metadata.name,
```
spwoodcock, line 51: Is would work here for example and all vars that match like this

nrjadkry (approved):

PR #74, opened 2024-07-17
feat: add gdal  & fix: cluster name from secrets

nischalstha9:


nrjadkry (approved):

spwoodcock:
Not saying it's wrong to add, but I'm just wondering what GDAL is needed for?

It's powerful but should be used wisely!
😃 (complicates running locally + adds 500MB or so to image size)

nrjadkry:
We need GDAL to read the raster files.  We need to read the raster files for generating flight plans with terrain following.

spwoodcock:
Oh ok, that works for now for testing. 

Future the DEMs should be entirely cloud based, so querying will be to an API rather than a file 👍

PR #76, opened 2024-07-18
feat: add contributor, task state, and outline to project detail API

Pradip-p:
### Description
This PR adds the following to the project detail API:
- Contributor information
- Task state
- Outline of each task
- Also remove short description

### Changes
- Updated `@router.get("/{project_id}", tags=["Projects"], response_model=project_schemas.ProjectOut)` to include the new fields.
- Updated schemas and relevant logic to handle the new data.

### DIFF
--- a/src/backend/app/projects/project_schemas.py
+++ b/src/backend/app/projects/project_schemas.py
@@ -1,8 +1,8 @@
 import uuid
-from pydantic import BaseModel, computed_field, Field
+from pydantic import BaseModel, computed_field, Field, validator
 from typing import Any, Optional, Union
 from geojson_pydantic import Feature, FeatureCollection, Polygon
-from app.models.enums import ProjectVisibility
+from app.models.enums import ProjectVisibility, State
 from shapely import wkb
 from datetime import date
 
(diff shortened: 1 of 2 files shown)


src/backend/app/projects/project_schemas.py
```diff
@@ -80,6 +78,17 @@ class TaskOut(BaseModel):
     id: uuid.UUID
     project_task_index: int
     outline: Any = Field(exclude=True)
+    state: Optional[State] = None
+    contributor: Optional[str] = None
+
+    @validator("state", pre=True, always=True)
```
nrjadkry, line 84: How do you get the State ?
state of the tasks are in `task_events`  table.

Pradip-p: @nrjadkry  To get the state, I am joining the tasks table with the task_events table, where the state of the tasks is stored.

```query = """
    SELECT
        tasks.id AS id,
        tasks.project_task_index AS project_task_index,
        tasks.outline AS outline,
        task_events.state AS state,
        users.name AS contributor
    FROM tasks
    LEFT JOIN task_events ON tasks.id = task_events.task_id
    LEFT JOIN users ON task_events.user_id = users.id
    WHERE tasks.project_id = :project_id;
"""
```

PR #79, opened 2024-07-18
hide add project button when signed in as drone operator

Prajwalism:


spwoodcock (approved):
This works fine for the current implementation 👍

But longer term I'm just wondering if we really need to use localStorage to manage the user role?

When we access the frontend we will be authenticating with the backend most likely, so it makes sense to just store the user role in the frontend state.

Plus surely the user role is tied to a specific project, meaning they could be a project manager in one project and a drone operator in another

PR #82, opened 2024-07-22
Update docker-entrypoint.sh

nischalstha9:
- Added pdm alembic migrations on entrypoint.sh


spwoodcock (approved):
Applying migrations on each application startup is generally bad practice, but is fine for this temp use case 👍

nrjadkry (approved):

PR #83, opened 2024-07-22
Feat/clipping no fly zones

Pradip-p:
### Description
This PR introduces enhancements to our geometry handling to support multiple Polygon and MultiPolygon types in the no_fly_zones field and also render the task split with no-fly zones.

### Changes
- Read and parse no-fly zones GeoJSON if provided.
- Calculate the difference between the project shape and no-fly zones.
- Generate the resulting GeoJSON feature with the new outline.

**Geometry Merging**:
- Implemented logic to merge multiple geometries into a single `MultiPolygon` using Shapely's `unary_union`.
- Ensured that if the merged geometry is not already a `MultiPolygon`, it is converted to one.

**screenshot**
multiPolygon
multi-polygon


spwoodcock (approved):
Via shapely is ok for now, but with PostGIS directly in the db would be a bit nicer 👍 

Note I recently tried to standardise our GeoJSON usage for FMTM, taking into account multipolygons, z-index, etc: https://github.com/hotosm/fmtm/blob/f1ec17040f1ab8819cc62129c19dae7e51ef72d2/src/backend/app/db/postgis_utils.py#L409

I'm thinking of pulling these utils out into a separate package `geojson-parsers` that could be used in FMTM & Drone-TM, so we could refactor in the future 😄

nrjadkry (approved):

PR #85, opened 2024-07-22
Upgrade Postgres v14 --> v16, plus migration for live server

### LINKED ISSUE #72: Use Postgres v16
## Issue

- Postgres 14 is 3yrs out of date at this point.
- https://github.com/hotosm/Drone-TM/blob/092316a3652733be1c1b2b5d6003889ed01d7446/docker-compose.yml#L38

## Solution

- We should update the docker-compose file to use Postgres v16 released last year.

    `image: postgis/postgis:16-3.4-alpine`

- Later this year Postgres v17 will be released and we can upgrade to that.

## Context

- Keeping on top of latest releases eases the maintenance burden / upgrade path going forward.
- This will also avoid us having to migrate v14 --> v16 if we start on v16.
- Would this affect anyone if the upgrade is done? Is there a db currently deployed anywhere?

spwoodcock:
Fixes #72

## Problem Solved

- Upgrading Postgres 14 --> 16, including a migration for the live server.

## Solution

- This was harder than I imagined!
- I made a separate docker-compose file to orchestrate the database upgrade.
- A script `contrib/pg-upgrade/upgrade-db.sh` can be used to automate the upgrade process.
- It checks if the upgrade has already taken place & is idempotent, so can be ran at every compose restart if required.
- See the comments in the compose file to see what each stage does.
- When redeploying to the AWS VM, it should upgrade automatically 🤞 

Let me know feedback / if anyone knows an easier way!
This is based on https://github.com/tianon/docker-postgres-upgrade

### DIFF
--- a/.github/workflows/build_and_deploy.yml
+++ b/.github/workflows/build_and_deploy.yml
@@ -87,8 +87,9 @@ jobs:
 
       - name: Deploy to VM
         run: |
-          docker compose --file docker-compose.vm.yml --env-file .env pull
+          bash contrib/pg-upgrade/upgrade-db.sh
+ 
           docker compose --file docker-compose.vm.yml --env-file .env up \
-            --detach --remove-orphans --force-recreate
+            --detach --remove-orphans --force-recreate --pull=always
         env:
           DOCKER_HOST: "ssh://${{ secrets.SSH_USER }}@${{ secrets.SSH_HOST }}"
--- a/contrib/pg-upgrade/Dockerfile
+++ b/contrib/pg-upgrade/Dockerfile
@@ -1,0 +1,9 @@
+FROM tianon/postgres-upgrade:14-to-16
+
+RUN set -ex \
+    && apt-get update \
+    && DEBIAN_FRONTEND=noninteractive apt-get install \
+    -y --no-install-recommends \
+        "postgresql-14-postgis-3" \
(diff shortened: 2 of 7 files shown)


mahesh-naxa:
@nischalstha9 can you look at these. I did go through it, using docker volumes & ability to upgrade db, LGTM. I am leaving you to test it out. 
thanks.

nischalstha9:
Worked like a charm at my side! Really a clever way to perform continuous upgrade! 👍

mahesh-naxa (approved):

sijandh35 (approved):

spwoodcock:
The proof is in the pudding as we say!

Let's see if it works on prod 😆

spwoodcock:
Forgot to make the pg-upgrade package public before deploy!

I made it public and updated to `ghcr.io/hotosm/drone-tm/pg-upgrade:14-to-16`

Workflow https://github.com/hotosm/Drone-TM/actions/runs/10058359643 🤞

nrjadkry:
@spwoodcock new deployments have failed.
Is this because of these changes ? 
https://github.com/hotosm/drone-tm/actions/runs/10071074857

spwoodcock:
Thanks, we are discussing here: https://github.com/hotosm/drone-tm/pull/86

PR #89, opened 2024-07-24
ci: added build timeout for infinitely hanging build steps

nischalstha9:
- Added timeout for workflow stage so that workflow won't keep running for unusual period of time.


nischalstha9:
#### Link that relates to hanging workflow:
https://github.com/hotosm/drone-tm/actions/runs/10071074857
https://github.com/hotosm/drone-tm/actions/runs/10069206358
https://github.com/hotosm/drone-tm/actions/runs/10062750584

image

mahesh-naxa (approved):
yeah we should definitely set some boundaries, looks good. 👍

PR #95, opened 2024-07-25
Feat/requested tasks

Pradip-p:
## Description

This PR adds functionality to send email notifications when a mapping request is either approved or rejected. It also includes an endpoint to fetch pending tasks for a specific project and user. The email template  has been updated to dynamically handle both scenarios using conditional logic.

## Changes

- **Email Notifications:**
  - Updated `event_handler` to include logic for sending approval and rejection email notifications.
  - Modified email template  to handle both approval and rejection scenarios.
  - Passed necessary context variables from the code to the template to dynamically adjust the content.
- **Pending Tasks Endpoint:**
  - Added `/requested_tasks/{project_id}/pending` endpoint to fetch a list of pending tasks for a specific project and user.


spwoodcock (approved):
Nice work!

Just to clarify for my understanding, will this send an email every time a task is locked for mapping? (the `MAP` event)

nrjadkry:
Yes, @spwoodcock An email to the project creator will be send when someone requests to lock it, and when accepted, another email to the drone operator will be send

nrjadkry (approved):

spwoodcock:
Damm, that's a lot of emails and red tape.

I guess it was a requirement!

Otherwise we could just notify the project manager, but not prevent the user from starting the task.

The task could be auto-unlocked after a certain timeout time

PR #102, opened 2024-07-29
feat: add endpoint to list all tasks for drone user with role validation

Pradip-p:
## Description
This PR adds a new endpoint to list all tasks for drone users. It includes role validation to ensure only users with the `DRONE_PILOT` role can access the endpoint. Additionally, the `get_tasks_by_user` function has been updated to calculate the task area and render task data using the `UserTasksStatsOut` Pydantic model.

### Key Changes
- Added endpoint to list all tasks for drone users with role validation.
- Implemented task area calculation in the `get_tasks_by_user` function.
- Rendered task data using `UserTasksStatsOut` Pydantic model with fields:
  - `task_id`
  - `task_area`
  - `created_at`
  - `state`

### DIFF
--- a/src/backend/app/tasks/task_routes.py
+++ b/src/backend/app/tasks/task_routes.py
@@ -1,7 +1,7 @@
 import uuid
 from fastapi import APIRouter, BackgroundTasks, Depends, HTTPException
 from app.config import settings
-from app.models.enums import EventType, State
+from app.models.enums import EventType, State, UserRole
 from app.tasks import task_schemas, task_crud
 from app.users.user_deps import login_required
 from app.users.user_schemas import AuthUser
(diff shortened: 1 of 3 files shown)


src/backend/app/tasks/task_routes.py
```diff
@@ -19,6 +19,27 @@
...
+):
+    """Get all tasks for a drone user."""
+
+    user_id = user_data.id
+    query = """SELECT role FROM user_profile WHERE user_id = :user_id"""
+    record = await db.fetch_one(query, {"user_id": user_id})
+    if not record:
+        raise HTTPException(status_code=404, detail="User profile not found")
+
+    if record.role != UserRole.DRONE_PILOT.name:
```
nrjadkry: I don't think we need this. Even Project creators should be allowed to map the tasks.

Pradip-p: @nrjadkry Thanks for the clarification. I’ve updated the code accordingly.

nrjadkry (approved):

PR #107, opened 2024-07-30
fix: index.html download always on restart & rm migration from entrypoint script

nischalstha9:
1. Fixes frontend ci not updating index.html
2. Removes migration cmd from entrypoint as we already have different service for it


spwoodcock (approved):
I haven't looked into the setup to really understand why the index.html is required in the backend, but the changes in this PR look fine 👍

Will review the implementation in more detail in future

PR #115, opened 2024-08-01
Upgrade backend to Python 3.11 + fix template dir permissions

spwoodcock:
Small PR to:
- Upgrade image to Python 3.11 + dependencies lock.
- The backend was failing to build due to `root` ownership of `src/backend/templates` dir. I set the permissions to 777 in the entrypoint script.
- Also adds PR labeller workflow / config for automatic labeling.

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -1,0 +1,16 @@
+"frontend":
+  - "src/frontend/**/*"
+"backend":
+  - "src/backend/**/*"
+"devops":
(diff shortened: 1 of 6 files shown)


nischalstha9 (approved):
Currently we are serving frontend index.html via backend so, we are using curl to download index.html file into backend from minio.

As mentioned on [PR 107](https://github.com/hotosm/drone-tm/pull/107), I think we should look into better ways to do it in future.

PR #125, opened 2024-08-06
Introduce Final Output and Auto Lock Tasks Fields to Improve Task Management

Pradip-p:
#### Summary:
This PR introduces two new fields, `final_output` and `auto_lock_tasks`, to the `DbProject` model and addresses issues with specific API endpoints. The `final_output` field allows for storing the desired output format for a project, enhancing flexibility and clarity in project configurations. The `auto_lock_tasks` field provides project creators with the option to enable or disable automatic task locking, giving more control over task management.

#### 1. Added `final_output` Field:
 - Introduced the `final_output` field in the `DbProject` model to store the output format (To store the desired output format for a projec).
#### 2. Added `auto_lock_tasks` Field:
 - Introduced the `auto_lock_tasks` field to define whether task auto-locking is optional for the project creator.

#### 1. Fixed Delete Project Endpoint:
- **Endpoint**: `/api/projects/{project_id}`
- Resolved issues to ensure projects can be deleted correctly.
#### 2. Fixed Get Pending Requested Tasks Endpoint:
- **Endpoint**: `/api/tasks/requested_tasks/pending`
- Addressed issues to ensure it correctly retrieves pending tasks.


src/backend/app/db/db_models.py
```diff
@@ -129,7 +130,10 @@ class DbProject(Base):
         ),
     )
     author = relationship(DbUser, uselist=False, backref="user")
-
+    final_output = cast(
```
nrjadkry: this is the multiselect option.  We might need an array field ?

Pradip-p: I’ve updated the schema & field to use an array field to better handle multiple selections.

src/backend/app/db/db_models.py
```diff
@@ -129,7 +130,8 @@ class DbProject(Base):
         ),
     )
     author = relationship(DbUser, uselist=False, backref="user")
-
+    final_output = cast(list, Column(ARRAY(Enum(FinalOutput))))
+    auto_lock_tasks = cast(bool, Column(Boolean, default=False))
```
nrjadkry: It is better to rename this `auto_lock_tasks` to something like  `requires_approval_from_manager_for_locking`.
Fields should be named clearly.

nrjadkry (approved):

PR #136, opened 2024-08-09
Refactor user module to use psycopg/pydantic

Pradip-p:
**Description:**

This PR refactors the application to use `psycopg` for database operations, replacing the previous encode/database setup. Key changes include:

- **Updated SQL Placeholders:** Converted SQL placeholders to `%(placeholders)s` to align with psycopg’s syntax.
- **Revised Routes:** Updated route handlers to utilize the new `user_deps.py` for CRUD operations and replacing the old `user_crud.py`.
- **Removed Deprecated Methods:** Cleaned up old methods and dependencies that are no longer used.
- **Testing:** Verified that all database interactions and endpoints function correctly with the new setup.


spwoodcock (approved):
Good work 👍

But just to clarify, the `_deps.py` file is only specifically to store dependency functions, that are used in Depends on the routes.

We should still use the `_crud.py` file to store the actual logic, used like `var = await user_crud.function_name()`

I mentioned that at some point we should probably rename all these files to `_logic.py` as it makes more sense, but we can wait to do that all together

nrjadkry (approved):

PR #137, opened 2024-08-09
Refactor Drone Module using psycopg and Pydantic 

nrjadkry:
Continuing this PR: [drone-tm/pull/128](https://github.com/hotosm/drone-tm/pull/128)

Updated Drone Modules using psycopg by replacing encode/databases


spwoodcock (approved):
Very nice!

PR #141, opened 2024-08-12
Refactor task module using pyscopg and pydantic

Pradip-p:
Continuing this PR: https://github.com/hotosm/drone-tm/pull/128

Updated Task Modules using psycopg by replacing encode/databases


src/backend/app/tasks/task_routes.py
```diff
@@ -53,9 +54,8 @@ async def new_event(
 
     match detail.event:
         case EventType.REQUESTS:
-            # TODO: Combine the logic of `update_or_create_task_state` and `request_mapping` functions into a single function if possible. Will do later.
             project = await get_project_by_id(db, project_id)
```
spwoodcock: Could we possibly get the project from `project.deps.get_project_by_id`?
Using `Depends`?

The project_crud.get_project_by_id function can probably be deleted

spwoodcock (changes requested):
Small change, but otherwise looks excellent!

spwoodcock:
Just saw your latest changes!

The distinction between logic that belongs on the schemas and logic in the `crud` file is blurry.

I think we should:
- Put these functions directly on the model in schemas for sure: `one`, `all`, `create`, `update`, `delete` (possibly others we think of, like `create_or_update`, but that will do for now).
- Then the `xxx_crud.py` files can be renamed `xxx_logic.py` and we put any endpoint logic here, such as `request_mapping` for example.

src/backend/app/tasks/task_routes.py
Pradip-p: Yes, we can indeed get the project using project.deps.get_project_by_id with Depends. I’ve now removed the project_crud.get_project_by_id function as it is redundant. Thanks for pointing that out!

nrjadkry (approved):

PR #164, opened 2024-08-23
Drone Image Processing

nrjadkry:
This Pull Request includes an endoint from which we can process the drone images to create the orthophoto and other assets using OpenDroneMap (ODM).

### Steps Involved
- [x] Download the images from S3 in a folder
- [x] Pass the Images into the processing script using internal docker networking
- [ ] Upload the output assets into S3
- [ ] Update the orthophoto url, pointcloud urls, etc into S3.
- [ ] Delete the downloaded drone images from the folder (They are in the S3)


src/backend/app/config.py
```diff
@@ -104,6 +104,8 @@ def assemble_db_connection(cls, v: Optional[str], info: ValidationInfo) -> Any:
     EMAILS_FROM_EMAIL: Optional[EmailStr] = None
     EMAILS_FROM_NAME: Optional[str] = "Drone Tasking Manager"
 
+    NODE_ODM_INTERNAL_DOCKER_URL: Optional[str] = "http://odm-api:3000"
```
spwoodcock: This could just be NODE_ODM_URL, and will either be the internal url if on the same machine, or an external url if hosted externally 👍

src/backend/app/projects/image_processing.py
```diff
@@ -0,0 +1,79 @@
...
+
+    def upload_images(self):
+        for image_file in os.listdir(self.images_dir):
+            image_path = os.path.join(self.images_dir, image_file)
+            if os.path.isfile(image_path):
+                self.upload_image(image_path)
+                log.info(f"Uploaded {image_file} successfully.")
+
+    def upload_image(self, image_path):
+        url = f"{self.base_url}/task/new/upload/{self.task_uuid}"
```
spwoodcock, line 44: I haven't checked the API, but is there possibly and endpoint to upload a zip?

It might actually be more efficient to download all the images, zip, then upload in one go - worth assessing / checking !

src/backend/app/projects/project_crud.py
```diff
@@ -282,3 +286,42 @@ async def preview_split_by_square(boundary: str, meters: int):
...
+    )
+
+    # Construct the images folder path
+    images_folder = (
+        f"/tmp/{project_id}/{task_id}"  # NOTE: local path is set to /tmp/ folder.
+    )
+    project_name = "Testing Project"  # FIXME: Use real Project name here
+
+    # Process images concurrently
+    with ThreadPoolExecutor() as executor:
```
spwoodcock: I wouldn't recommend usage of the threadpool in an async framework like FastAPI.

This approach will work and could be fine for now, but it might choke the server under load as threads are overallocated.

It's much more resource efficient to process this asynchronously using an aiohttp connection pool, but it does mean installing something like aiohttp as a dependency.

Downloading multiple files this way only uses one thread. The downloads are added to a queue and the selector checks when the download is ready to be processed by the OS and opens a connection as needed. It's more efficient than having multiple threads open connections and keeping them open until done

spwoodcock (changes requested):
Looks good!

As discused, this is the best approach until later when we can make a PR for ODM processing direct from S3 😄

Some minor comments, but this would work as it is if needed 👍

PR #185, opened 2024-09-02
Feat: add download boundaries endpoint and auto-approve task locking for project owner

Pradip-p:
### Description:
This PR introduces the ability to dynamically download either the entire project AOI or individual task boundaries based on user input & the task locking process by automatically approving requests from the project owner. Previously, even if the project owner requested to lock a task, it would still require approval and be marked as pending. With this update, the system will now automatically approve and lock the task if the requesting user is the project owner.

- Added a new endpoint to download project task boundaries or AOI as a GeoJSON file.
- Updated the `Task` schemas with a static method `get_task_geometry` to handle fetching task or project geometries from the database.
- Incorporated logic to handle optional `task_id` and `split_area` parameters for selective downloads.
- Improved error handling for cases where task geometries or project outlines are not found.
- Set appropriate headers for GeoJSON file downloads, including content disposition and type.

- Automatic Approval for Task Locking by Project Owner: Automatically approves and locks tasks when the project owner requests it, bypassing the need for manual approval.

### screenshot
Screenshot from 2024-09-02 12-29-59


src/backend/app/projects/project_routes.py
```diff
@@ -16,13 +27,74 @@
...
+            description="The project ID in UUID format.",
+        ),
+    ],
+    db: Annotated[Connection, Depends(database.get_db)],
+    user_data: Annotated[AuthUser, Depends(login_required)],
+    task_id: Optional[UUID] = Query(
+        default=None,
+        description="The task ID in UUID format. If not provided, all tasks will be downloaded.",
+    ),
+    split_area: bool = Query(
```
nrjadkry, line 52: What does this do ?

Pradip-p: > What does this do ?

The `split_area` parameter is used to control the type of geometry returned in the downloaded GeoJSON file:

- **When `split_area` is `True`:**  
  The GeoJSON file will contain the boundaries of individual tasks within a project. These tasks are typically smaller subdivisions of the overall project area, and each one will be represented as a separate polygon. The resulting GeoJSON will be a `FeatureCollection` where each feature corresponds to a task boundary.
Screenshot from 2024-09-03 13-12-50

- **When `split_area` is `False`:**  
  The GeoJSON file will contain the entire Area of Interest (AOI) as a single geometry. 
Screenshot from 2024-09-03 13-12-11

nrjadkry (approved):

PR #192, opened 2024-09-04
 Refactor: remove fmtm-splitter and Update Profile Endpoints

Pradip-p:
### Description:
- Removed `fmtm-splitter`: Eliminated the `fmtm-splitter` dependency; the project now uses a separate `TaskSplitter` algorithm for area splitting.
- Profile Update Endpoint: Added response messages for successful profile updates.
- Enhanced my-info Endpoint: Expanded to display all available user profile information and validated using a Pydantic class.
- Dynamic Email Notification URL: Loaded `FRONTEND_URL` from environment files and set a dynamic URL for  click button.


src/backend/app/email_templates/mapping/requests.html
```diff
@@ -130,7 +130,7 @@ <h2>Mapping Task Details</h2>
             <p><strong>Project:</strong>{{project_name}}</p>
             <p><strong>Description:</strong> {{description}}</p>
           </div>
-          <a href="dronetm.naxa.com.np" class="task-button"
+          <a href="https://dronetm.org/" class="task-button"
```
nrjadkry: We should take a base url from env.

Pradip-p: > We should take a base url from env.

@nrjadkry Good point! I've updated the code to load the base URL from the environment variables.

src/backend/app/config.py
```diff
@@ -80,6 +80,7 @@ def assemble_db_connection(cls, v: Optional[str], info: ValidationInfo) -> Any:
         )
         return pg_url
 
+    SITE_URL: str = "http://localhost:3040"
```
nrjadkry: Better to name it FRONTEND_URL ?

nrjadkry (approved):

PR #200, opened 2024-09-09
Feat: Add optional filter for projects by current user

Pradip-p:
### Description:

This pull request introduces a new feature to optionally filter projects based on the current user. 

### Changes:

- Added `filter_by_owner` to the `read_projects` endpoint to filter projects where the `author_id` matches the current user's ID. If `filter_by_owner` is set to `True`, the query will only return projects authored by the current user. If `filter_by_owner` is set to `False` or not provided, all projects will be returned.

### DIFF
--- a/src/backend/app/projects/project_schemas.py
+++ b/src/backend/app/projects/project_schemas.py
@@ -278,7 +278,7 @@ class DbProject(BaseModel):
 
     async def all(
         db: Connection,
-        user_id: Optional[uuid.UUID] = None,
+        user_id: Optional[str] = None,
         skip: int = 0,
         limit: int = 100,
     ):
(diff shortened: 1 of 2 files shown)


src/backend/app/projects/project_schemas.py
```diff
@@ -278,20 +278,27 @@ async def one(db: Connection, project_id: uuid.UUID):
 
     async def all(
         db: Connection,
+        user_id: Optional[uuid.UUID] = None,
```
nrjadkry: is the user id UUID? If not, better to have it `str`

Pradip-p: @nrjadkry  I checked and found that `user_id` is stored as `str`. I've updated the code to reflect this.

nrjadkry (approved):

PR #227, opened 2024-09-19
Drone Image Processing using Open Drone Map (ODM)

nrjadkry:
## Overview

This pull request introduces a comprehensive image processing pipeline that:

- [x] Downloads images from MinIO (S3-compatible storage).
- [x] Processes the images using NodeODM (via PyODM).
- [x] Uploads the processed results as a ZIP file back to MinIO.
- [x] Integrates the processing into a FastAPI application using background tasks.
- [x] Provides endpoints to initiate processing and check task status.

## Key Features
DroneImageProcessor Class: Manages the entire processing workflow, including downloading images, processing with NodeODM, and uploading results.

S3 Integration: Uses existing functions in `s3.py` to download the images and upload the processed imagery back to MinIO

FastAPI Background Tasks: Implements background processing to keep the API responsive.

Task Status Tracking: Introduces a TaskManager to track the status of processing tasks.

API Endpoints:
POST `process_imagery/{project_id}/{task_id}/` Initiates processing

### DIFF
--- a/src/backend/app/projects/image_processing.py
+++ b/src/backend/app/projects/image_processing.py
@@ -3,7 +3,7 @@ import uuid
 import tempfile
 import shutil
 from pyodm import Node
-from app.s3 import get_file_from_bucket, list_objects_from_bucket
+from app.s3 import get_file_from_bucket, list_objects_from_bucket, add_file_to_bucket
 from loguru import logger as log
 from concurrent.futures import ThreadPoolExecutor
 
@@ -113,8 +113,9 @@ class DroneImageProcessor:
         :param output_path: The directory where results will be saved.
         """
         log.info(f"Downloading results to {output_path}...")
-        task.download_assets(output_path)
+        path = task.download_zip(output_path)
         log.info("Download completed.")
+        return path
 
     def process_images_from_s3(self, bucket_name, name=None, options=[]):
         """
@@ -130,9 +131,7 @@ class DroneImageProcessor:
         # Create a temporary directory to store downloaded images
         temp_dir = tempfile.mkdtemp()
         try:
-            self.download_images_from_minio(
-                bucket_name, self.project_id, self.task_id, temp_dir
-            )
+            self.download_images_from_minio(bucket_name, temp_dir)
 
             images_list = self.list_images(temp_dir)
 
@@ -140,9 +139,16 @@ class DroneImageProcessor:
             task = self.process_new_task(images_list, name=name, options=options)
             # Monitor task progress
             self.monitor_task(task)
+
             # Optionally, download results
-            self.download_results(task, output_path="/tmp/{self.project_id}/assets")
+            output_file_path = f"/tmp/{self.project_id}"
+            path_to_download = self.download_results(task, output_path=output_file_path)
+
+            # Upload the results into s3
+            s3_path = f"projects/{self.project_id}/{self.task_id}/assets"
+            add_file_to_bucket(bucket_name, path_to_download, s3_path)
             return task
+
         finally:
             # Clean up temporary directory
             shutil.rmtree(temp_dir)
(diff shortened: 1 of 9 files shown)


src/backend/app/projects/image_processing.py
```diff
@@ -0,0 +1,155 @@
...
+        if options is not None:
+            for o in options:
+                opts[o["name"]] = o["value"]
+        return opts
+
+    def download_object(self, bucket_name: str, obj, images_folder: str):
+        if obj.object_name.endswith((".jpg", ".jpeg", ".JPG", ".png", ".PNG")):
+            # log.info(f"Downloading image from s3 {obj.object_name}")
+            local_path = f"{images_folder}/{os.path.basename(obj.object_name)}"
+            os.makedirs(os.path.dirname(local_path), exist_ok=True)
```
spwoodcock: This will work fine, but I would always recommend using Pathlib 👍

```python
from pathlib import Path
Path(local_path).mkdir(exist_ok=True)
```

src/backend/app/projects/image_processing.py
```diff
@@ -0,0 +1,155 @@
...
+            get_file_from_bucket(bucket_name, obj.object_name, local_path)
+
+    def download_images_from_minio(self, bucket_name, local_dir):
```
spwoodcock: Remember that we could be using any S3 compatible storage.

We just happen to use Minio, but AWS S3 would work just as well.

Keep this in mind when naming functions and variables + writing docstrings 👍

src/backend/app/projects/image_processing.py
```diff
@@ -0,0 +1,155 @@
...
+
+    def list_images(self, directory):
+        """
+        Lists all images in the specified directory.
+
+        :param directory: The directory containing the images.
+        :return: List of image file paths.
+        """
+        images = []
+        for root, dirs, files in os.walk(directory):
```
spwoodcock: A pathlib.Path.glob generator could do the same here

src/backend/app/projects/image_processing.py
```diff
@@ -0,0 +1,155 @@
...
+        :param task_id: The task UUID.
+        :param local_dir: Local directory to save the images.
+        :return: List of local image file paths.
+        """
+        prefix = f"projects/{self.project_id}/{self.task_id}"
+
+        objects = list_objects_from_bucket(bucket_name, prefix)
+
+        # Process images concurrently
+        with ThreadPoolExecutor() as executor:
```
spwoodcock, line 58: I would recommend asyncio over threading for this type of downloading onto a filesystem.

`asyncio.gather` would work nicely and be less blocking (using a single thread via coroutines).

Example from FMTM:

https://github.com/hotosm/fmtm/blob/35e7240cc78d62c288f0ea60aef0eb7bde8beaed/src/backend/app/central/central_crud.py#L560-L570

But not this would mean you need to update functions to `async` and use `await`

src/backend/app/waypoints/waypoint_routes.py
```diff
@@ -105,16 +105,19 @@ async def get_task_waypoint(
 
     if project.is_terrain_follow:
         dem_path = f"/tmp/{uuid.uuid4()}/dem.tif"
-        get_file_from_bucket(
-            settings.S3_BUCKET_NAME, f"projects/{project_id}/dem.tif", dem_path
-        )
+        try:
+            get_file_from_bucket(
```
spwoodcock, line 109: As you rightly say below, once everything is working and we are happy, this definitely needs to be changed getting elevation data by querying the cloud optimised DEM. A DEM can be quite large, so downloading every time we process imagery is very inefficient

spwoodcock (approved):
Nice work 🙌

This should function is it is, but I left some comments for optimisation if needed.

There are many areas we can optimise for sure, but it's good to get a first version working before over-optimising 😁

PR #317, opened 2024-10-29
Relock Python dependencies on the backend

spwoodcock:
- The pdm.lock was quite outdated and hadn't been relocked since version updates in pyproject.toml.
- As a result the backend dockerfile could not be built.
- I relocked the deps to fix this.


spwoodcock:
Ah, I was using an old version of the repo where `main` was the default branch!
`develop` is actually fine and has a recent lock file 👍 
Closing.

PR #324, opened 2024-11-05
fix(s3-config): update S3 bucket name & path prefix to mimic with dev & prd

Pradip-p:


src/backend/app/config.py
```diff
@@ -86,7 +86,8 @@ def assemble_db_connection(cls, v: Optional[str], info: ValidationInfo) -> Any:
     S3_ENDPOINT: str = "http://s3:9000"
     S3_ACCESS_KEY: Optional[str] = ""
     S3_SECRET_KEY: Optional[str] = ""
-    S3_BUCKET_NAME: str = "dtm-data"
+    S3_BUCKET_NAME: str = "dronetm"
```
nrjadkry: We can change this from environment variable

PR #343, opened 2024-11-19
Upload align.laz along with Drone Images

smathermather:
## Feature: Upload align.laz with drone images

This pull request introduces a new feature to upload an align.laz file. The file should be with the name `align.laz` and should be within the same folder as the images.

This will be critical to the workflow to ensure alignment of multiple missions, similar to multi-temporal example here:
https://docs.opendronemap.org/map-accuracy/#multi-temporal-datasets


nrjadkry (approved):
Thanks @smathermather

PR #346, opened 2024-11-20
Upload task orthophoto to OpenAerialMap (OAM)

Pradip-p:
## Description:
This PR introduces a new endpoint for uploading an orthophoto (TIFF file) to OpenAerialMap (OAM).

- [x] Implemented a POST endpoint at /upload/{project_id}/{task_id} for uploading orthophotos (TIFF files) to OpenAerialMap.
- [x]  Added functionality to generate S3 URL for the orthophoto and initiate the upload to OAM using the OAM API token.
- [x] Returns OAM upload ID for tracking the upload status.


src/backend/app/tasks/task_routes.py
```diff
@@ -86,3 +89,43 @@ async def new_event(
...
+    user_data: Annotated[AuthUser, Depends(login_required)],
+    project: Annotated[
+        project_schemas.DbProject, Depends(project_deps.get_project_by_id)
+    ],
+):
+    """
+    Uploads an orthophoto (TIFF) file to OpenAerialMap (OAM).
+
+    Args:
+        project_id: The UUID of the project.
```
spwoodcock, line 107: You only really need to add docstrings with `Args` and `Returns` on crud functions. For routes, this is automatically documented using type hints and OpenAPI spec 👌 

Plus docstrings are useful for **devs** to consume functions, but the route is the top level of the logic and nobody is importing and using the routes elsewhere

src/backend/app/tasks/task_routes.py
```diff
@@ -86,3 +89,43 @@ async def new_event(
         user_data,
         background_tasks,
     )
+
+
+@router.post("/upload/{project_id}/{task_id}")
```
spwoodcock, line 94: This probably isn't a comment specific to this PR, but the project as a whole.

The endpoint structure for FMTM doesn't necessarily conform to the REST spec / best practices, but we are working on it & hopefully that's changing over time 🙏 

So typically it would be set up like this:

If the task id is a unique identifier on it's own.
`/projects/{project_id}`
`/projects/{project_id}/do-something`
`/tasks/{task_id}`
`/tasks/{task_id}/do-something`

If task ids are only unique in relation to the project they are within:
`/projects/{project_id}`
`/projects/{project_id}/tasks/{task_id}`
`/projects/{project_id}/tasks/{task_id}/do-something`

So for this endpoint, as task_id is a globally unique UUID, it would make sense to create:

`/tasks/{task_id}/upload-ortho`

or something similar.

Just for info going forward! I'm not saying things have to change immediately, or even at all, but it could be a nice to have into the future 😄

spwoodcock (commented):
Looking good!

nrjadkry:
I think we can add some tags into OpenAerialMap. We can add some default tags like `#drone-tm` and ask the users  for additional tags when uploading to OAM.

spwoodcock:
> I think we can add some tags into OpenAerialMap. We can add some default tags like `#drone-tm` and ask the users  for additional tags when uploading to OAM.
> 
> And also instead of uploading the imagery from our account, it might be better to ask the users for their own token .  What do you say? @ivangayton @spwoodcock 
> 

Definitely! Add #drone-tm and something like #{domain_name}-{project-id} as a reference to start.

Also agree with Ivan to attribute the project manager who clicks the upload button.

Can you use the oauth flow for OAM to log in? Then I assume you can use the oauth token from OAM to upload as a specific user.
Or perhaps they have an endpoint to generate API keys, so you could generate a key, possibly store in our db, then use that.
A few options that warrant investigation 👍

PR #347, opened 2024-11-21
Replace `pdm` to `uv` for faster dependency management

Pradip-p:
Replaced `pdm` with `uv` for faster dependency management. `uv` is written in Rust, making it much faster for locking dependencies and managing Python versions.

#### Changes:
- Switched all `pdm` usage to `uv`.
- Refined Dockerfile for `uv` compatibility.
- Added pre-commit hook to auto-update the lockfile.

#### Bonus:
Pre-commit hook ensures lockfile updates automatically.

#### Reference:
Inspired by [[fmtm PR #1891](https://github.com/hotosm/fmtm/pull/1891)](https://github.com/hotosm/fmtm/pull/1891).

@spwoodcock   @nrjadkry  @nischalstha9 , I've updated the project to use `uv` as the package manager. Could you please review the changes.


src/backend/Dockerfile
```diff
@@ -22,19 +22,43 @@ RUN set -ex \
...
 
-# Install PDM
-RUN pip install -U pdm
-# Disable update check
-ENV PDM_CHECK_UPDATE=false
-
 WORKDIR /project
 
+# Install uv
+COPY --from=ghcr.io/astral-sh/uv:0.5.2 /uv /bin/uv
```
spwoodcock, line 28: Does this definitely work? I'm sure you tested it and I just didn't realise this syntax was possible 😅 

But I thought that `--from` and external source never used to be an option!

That's why I used this pattern:
```dockerfile
FROM ghcr.io/astral-sh/uv:0.5.2 as uv-image
...
COPY --from=uv-image /uv /bin/uv
```

src/backend/Dockerfile
```diff
@@ -22,19 +22,43 @@ RUN set -ex \
...
+# Install dependencies using uv (caching to improve speed for repeated builds)
+RUN --mount=type=cache,target=/root/.cache/uv \
+    --mount=type=bind,source=pyproject.toml,target=pyproject.toml \
+    --mount=type=bind,source=uv.lock,target=uv.lock \
+    uv sync --frozen --no-install-project
 
-# Install dependencies and project into the local packages directory
-RUN pdm install --check --prod --no-editable
+# Sync the project again using uv, ensuring all project dependencies are installed
+RUN --mount=type=cache,target=/root/.cache/uv \
```
spwoodcock, line 60: What's the reason for the double `sync`?

spwoodcock (commented):
Nice change!

src/backend/Dockerfile
Pradip-p: > What's the reason for the double `sync`?

```# Install dependencies using uv (caching to improve speed for repeated builds)
RUN --mount=type=cache,target=/root/.cache/uv \
    --mount=type=bind,source=pyproject.toml,target=pyproject.toml \
    --mount=type=bind,source=uv.lock,target=uv.lock \
    uv sync --frozen --no-install-project

# Sync the project again using uv, ensuring all project dependencies are installed
RUN --mount=type=cache,target=/root/.cache/uv \
    uv sync
```
I found, the first uv sync with --no-install-project caches dependencies, while the second installs the project after copying the code. This speeds up builds. Details: [[link](https://docs.astral.sh/uv/guides/integration/docker/#intermediate-layers)].

spwoodcock: Love the idea!

But the code is not copied in between the two run commands, meaning this extra step doesn't really do anything 😆

(You could probably remove the second sync with no consequence)

src/backend/Dockerfile
spwoodcock: Nice! Thus must have been added in a recent docker version 
https://docs.docker.com/build/building/multi-stage/#use-an-external-image-as-a-stage

spwoodcock (approved):
Test out removing that extra sync stage to see if it's needed, otherwise all good 👍

nrjadkry (approved):

PR #464, opened 2025-02-03
Add matomo cookie / tracking acceptance banner

spwoodcock:

- [x] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation
- [ ] 🧑‍💻 Refactor
- [ ] ✅ Test
- [ ] 🤖 Build or CI

## Related Issue

Anticipates merge of #463
the only change here is actually to App.tsx.

## Describe this PR

- Adds matomo cookie tracking banner as requested.
- Accepting will enable anonymised matomo tracking.
- Rejecting will prevent any tracking.
- Both will close the banner, with response saved to localStorage.

> [!NOTE]
> I had to do a hacky workaround using `style={{position: 'fixed', bottom: '0%'}}` for fixed position on the screen.
>
> This is due to https://github.com/hotosm/ui/issues/72, so could be removed once fixed.

## Screenshots

image


emi420:
Do you think is necessary to have tracking with cookies? we could have visit metrics without cookies, which means that we wouldn't need the annoying "About the information we collect" message.  

This would do it:

```<!-- Matomo -->
<script>
  	var _paq = window._paq = window._paq || [];
  	/* tracker methods like "setCustomDimension" should be called before "trackPageView" */
        _paq.push(['disableCookies']);
  	_paq.push(['trackPageView']);
  	(function() {
   	 var u="//matomo.hotosm.org/";
   	 _paq.push(['setTrackerUrl', u+'matomo.php']);
   	 _paq.push(['setSiteId', '35']);
   	 var d=document, g=d.createElement('script'), s=d.getElementsByTagName('script')[0];
   	 g.async=true; g.src=u+'matomo.js'; s.parentNode.insertBefore(g,s);
	  })();
</script>
<!-- End Matomo Code -->
```

spwoodcock:
Also as a side note, are you advocating for each project to insert its own matomo script, instead of standardising it in the web component?

I sense a reluctance to use the components, so just want to understand any reasoning 😃

emi420:
You don't need the banner for authentication, the consent is needed for cookies that are not strictly necessary (ref: https://europa.eu/youreurope/business/dealing-with-customers/data-protection/online-privacy/index_en.htm).

I don't have an opinion on how the script should be inserted, we need it for obtaining metrics about usage of tools (ex: number of unique visitors) it could be trough a web component if you think that's the best approach.

PR #473, opened 2025-02-14
Created custom permission classes and implementation in api endpoints

nrjadkry:
Different custom permission classes have been created.
Some of them are :
- [x] IsAuthenticated
- [x] IsSuperUser
- [x] IsProjectCreator

`check_permissions`  dependency function is created that handles all the permissions classes passed into this function from routes and checks the relevant permissions.

An example of such endpoints looks like this:
```
@router.delete("/{project_id}", tags=["Projects"])
async def delete_project_by_id(
    db: Annotated[Connection, Depends(database.get_db)],
    project: Annotated[
        project_schemas.DbProject,
        Depends(
            check_permissions(
                IsSuperUser() | IsProjectCreator(),
                get_obj=project_deps.get_project_by_id,
            )
        ),
    ],
):
    project_id = await project_schemas.DbProject.delete(db, project.id)
    return {"message": f"Project successfully deleted {project_id}"}
```

In this endpoint, 2 permissions are checked, IsSuperUser and IsProjectCreator. OR Operator is used and the project can be deleted when either the user is a super user of the project creator.

@spwoodcock What do you think about checking the user permissions in this approach. Do you have any other alternative ideas ? I wanted to make the generic permission classes that can be used across the project.

### DIFF
--- a/src/backend/app/projects/project_routes.py
+++ b/src/backend/app/projects/project_routes.py
@@ -39,9 +39,13 @@ from app.utils import (
     timestamp,
     send_project_approval_email_to_regulator,
 )
-from app.users import user_schemas
 from app.jaxa.upload_dem import upload_dem_file
 from minio.deleteobjects import DeleteObject
+from app.users.permissions import (
+    IsProjectCreator,
+    IsSuperUser,
+    check_permissions,
+)
 
 
 router = APIRouter(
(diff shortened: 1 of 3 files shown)


spwoodcock:
This is pretty neat if you need a granular permission approach 👍

But as far as I know, the permissions in DroneTM are heirachical like FMTM right?

super admin
org admin
project admin
field admin?
operator

The roles go in defending order of access, meaning if you specify:

- org admin --> org admin and super admin have access

- operator --> all above roles have access 

It seems a bit redundant to have to list every role that has access when you have this type of approach IMO 😄

nrjadkry:
You're right that a hierarchical permission system can simplify access control, but in this project, the roles are structured a bit differently than in FMTM. Specifically:

1. **Organizations** aren't present (though they might be introduced in the future).
2.  **The core roles** are: Super Admin, Project Creator, and Drone Operator.

I am using IsAuthenticated role to every other endpoint. Other endpoints that requires specific roles like project creator and super admin checks just those roles.

PR #488, opened 2025-02-21
Setup ARQ Background Task System

nrjadkry:
# Add ARQ Background Task System

## Description
This PR implements a background task system using ARQ (Async Redis Queue) . It includes a test endpoint to demonstrate task queuing

## Changes Made
- Added ARQ worker configuration with proper Redis connection handling
- Implemented sleep_task as a demonstration background task
- Added test endpoint for task enqueueing
- Added queue management with default queue configuration


spwoodcock (commented):
Could be good 👍
According to the maintainer arq is going to get some more love as part of the astral group now.

What tasks do we need the queue for?

nrjadkry:
> Could be good 👍 According to the maintainer arq is going to get some more love as part of the astral group now.
> 
> What tasks do we need the queue for?

Fow now, image processing is the first even though it is running most of its tasks in async event loop.

> Could be good 👍 According to the maintainer arq is going to get some more love as part of the astral group now.
> 
> What tasks do we need the queue for?

Right now, image processing is the first task for the queue. While most of the processing runs in the async event loop, tracking tasks has become quite challenging—especially when uploading images to ODM.

spwoodcock:
Arq is a full task queue and runner, but it seems we just need to keep track of processing in external systems right?

Wouldn't a simple db table do this? `external_jobs` table with status and task id?

nrjadkry:
Tracking tasks solely with a database table didn't resolve all our issues. Uploading images to ODM isn't asynchronous, leading to bottlenecks when handling multiple requests simultaneously. Therefore, I opted to implement a queue

spwoodcock:
Got you - arq is a good temp solution.

The downloading of imagery from S3 to reupload to ODM was always a temp workaround.

In the long run ODM should be downloading the imagery from S3 itself, for processing.

Apologies if I have forgotten, but did we ever attempt this approach for downloading via zip URL and then using a webhook on completion?

https://github.com/OpenDroneMap/NodeODM/issues/181#issuecomment-2411278354

nrjadkry:
Since we primarily use AWS S3 for storing images, I couldn't find a way to download the images in a single zip url.

Do you know about any way to do that?

But, we can get the zip_url from minio. I did not find it in AWS

spwoodcock:
> Since we primarily use AWS S3 for storing images, I couldn't find a way to download the images in a single zip url.
> 
> Do you know about any way to do that?
> 
> But, we can get the zip_url from minio. I did not find it in AWS

Minio has the functionality in built, but for AWS S3 this is a good use case for a lambda function.

Stream multiple files to a single zip. Example:

https://amiantos.net/zip-multiple-files-on-aws-s3/

docker-compose.yml
```diff
@@ -109,3 +109,30 @@ services:
...
+    depends_on:
+      - redis
+    env_file: .env
+    networks:
+      - dtm-network
+    volumes:
+      - ${PROJECT_DIR:-.}/src/backend:/project/src/backend
+      - frontend-html:/project/src/backend/frontend_html
+    restart: unless-stopped
+    healthcheck:
```
spwoodcock, line 127: Healthchecks are to check if the main service in the container is healthy and signal to the container orchestrator if it should restart.

So checking the db isn't the best thing to do here.

Instead there is an arq CLI helper for this:

```bash
arq app.arq.tasks.WorkerSettings --check
```

So you could set that as the CMD instead

src/backend/app/arq/health_check.py
```diff
@@ -0,0 +1,24 @@
+import asyncio
+from app.arq.tasks import WorkerSettings
+from arq.worker import check_health
+from psycopg_pool import AsyncConnectionPool
+from app.config import settings
+
+
+async def check_db_health():
```
spwoodcock: This file can be deleted 👍

spwoodcock (commented):
Looking good!

PR #630, opened 2025-10-07
[FIX]: Local Development Setup

FeezyHendrix:

- [ ] 🍕 Feature
- [x] 🐛 Bug Fix
- [ ] 📝 Documentation
- [ ] 🧑‍💻 Refactor
- [ ] ✅ Test
- [ ] 🤖 Build or CI

## Related Issue

Fixes local development setup issues preventing proper service connectivity.

## Describe this PR

Fixes critical local dev configuration issues: corrected NodeODM connection from `http://odm-api:3000` to `http://nodeodm:9900`, added `S3_DOWNLOAD_ROOT` for browser Minio access, aligned S3 bucket naming to `dtm-bucket`, and fixed Alembic migrations to use psycopg3 driver instead of psycopg2.

## Review Guide

1. Run `docker compose down && docker compose up -d`
2. Verify migrations complete successfully
3. Test image upload to Minio from frontend
4. Confirm NodeODM connectivity for processing tasks

- [x] 📖 Read the HOT Code of Conduct
- [x] 👷‍♀️ Create small PRs
- [x] 📝 Use descriptive commit messages
- [x] 📗 Update any related documentation and include any relevant screenshots


.env.example
```diff
@@ -8,7 +8,7 @@ POSTGRES_HOST=${POSTGRES_HOST:-db}
 S3_BUCKET_NAME=${S3_BUCKET_NAME:-dtm-bucket}
 S3_ACCESS_KEY=${S3_ACCESS_KEY:-SAMPLEACCESSKEYFORMINIOROOT}
 S3_SECRET_KEY=${S3_SECRET_KEY:-SAMPLESECRETACCESSKEYFORMINIOROOT}
-S3_DOWNLOAD_ROOT=${S3_DOWNLOAD_ROOT:-http://minio:9000}
```
spwoodcock, line 11: We need to use the internal minio URL for access from the API, e.g. for generating pre-signed URls, then then the external (in this case `localhost:9000`) URL for direct access from the frontend.

For example in FieldTM we configure like this:
https://github.com/hotosm/field-tm/blob/9fcf05cc8767600b87e3315d56fb07431133010f/src/backend/app/config.py#L331

It does cause complications in the pre-signed URL signatures though when developing on local dev, so we need an extra function to use in certain places with `if settings.DEBUG=True`:

```python
def strip_presigned_url_for_local_dev(url: str) -> str:
    """Helper for local development handling docker URL + pre-signing.

    For local dev only, we need to iterate through and replace S3_ENDPOINT
    with S3_DOWNLOAD_ROOT, due to internal docker name used for S3 URL.
    The bucket is also public in local dev, so we remove pre-signed portion
    of URL, giving us direct access without a signature mismatch
    """
    host_accessible_url = url.replace(settings.S3_ENDPOINT, settings.S3_DOWNLOAD_ROOT)
    try:
        split_url_on_presign_vars = host_accessible_url.split("?")
        return split_url_on_presign_vars[0]
    except Exception as e:
        log.debug(f"Failed to convert S3 URL for local development: {e}")
        return url
```

Does this change work as intended?

FeezyHendrix: Ah, I see! it does work as intended locally. Hence we can access the url directly. The helper function would work to replace the url sent to the frontend.

FeezyHendrix: @spwoodcock Done the change requested. 
Reverted the .env.example to use the minio internal instance, also added the util function.

spwoodcock:
Small comment, could we use conventional commit format?

https://docs.hotosm.org/dev-guide/repo-management/version-control/#conventional-commits

It helps automate changelog generation

PR #632, opened 2025-10-10
[Refactor]: Ensure Geojson is validated properly.

FeezyHendrix:

- [x] 🐛 Bug Fix
- [x] 🍕 Feature

## Related Issue

Fixes issue with LineString geometries causing backend error: "Couldn't merge the multipolygon to polygon: object of type 'float' has no len()"

## Describe this PR

This PR adds frontend validation to detect and convert LineString geometries to Polygon format before uploading GeoJSON files. 

**Problem:** Users could upload GeoJSON files containing LineString geometries, which caused the backend `merge_multipolygon` function to fail since it expects Polygon/MultiPolygon geometries only.

**Solution:** 
- Added `ensurePolygonGeometry()` utility function that validates LineString geometries
- Checks if LineString is properly closed (first point = last point)
- Automatically converts valid closed LineStrings to Polygons
- Shows error toast messages if LineString is invalid (not closed)
- Prevents invalid data from reaching the backend

## Screenshots

N/A - Backend error prevention

- Backend fix to handle LineString geometries - decided frontend validation is better UX
- Auto-closing LineStrings - rejected as this could hide user errors

## Review Guide

**To test:**
1. Create a GeoJSON file with a closed LineString geometry (first and last coordinates are identical)
2. Upload it in the "Define AOI" step when creating a project
3. Verify it's automatically converted to a Polygon and accepted
4. Create a GeoJSON with an unclosed LineString 
5. Upload it and verify you get an error toast: "LineString must be closed..."

**Example test files:**
- Valid (closed): LineString where `coordinates[0] === coordinates[last]`
- Invalid (unclosed): LineString where first ≠ last point

- [x] 📖 Read the HOT Code of Conduct: <https://docs.hotosm.org/code-of-conduct>
- [x] 👷‍♀️ Create small PRs. In most cases, this will be possible.
- [ ] ✅ Provide tests for your changes.
- [x] 📝 Use descriptive commit messages.
- [x] 📗 Update any related documentation and include any relevant screenshots.


src/frontend/src/utils/convertLayerUtils.ts
```diff
@@ -49,6 +49,66 @@ export function convertGeojsonToFile(geojson: Record<string, any> | null) {
...
+
+      // Check if LineString is closed (first and last points are the same)
+      const firstPoint = coords[0];
+      const lastPoint = coords[coords.length - 1];
+      const isClosed =
+        firstPoint[0] === lastPoint[0] && firstPoint[1] === lastPoint[1];
+
+      if (!isClosed) {
+        errors.push(
+          `Feature ${index + 1}: LineString must be closed (first and last points must be identical) to create a valid Polygon`
```
spwoodcock, line 77: This part doesn't really make sense.

A closed linestring does not equal a polygon.

We should only support uploading polygons, polygons wrapped in a feature, and possibly multipolygons 😅

spwoodcock (changes requested):
Thanks for the input!

So I'm not 100% sure this is ready yet, but could we attempt to integrate this package instead, with a small endpoint wrapper, to do all geojson validation?

https://github.com/hotosm/geojson-aoi-parser

Writing custom logic on the frontend probably isn't needed. Plus I'm not sure you have used geojson much before, so no need to try and work out all the edge cases yourself 👍

Could be a good first little task that is backend and frontend 🙏

src/frontend/src/utils/convertLayerUtils.ts
FeezyHendrix: My thought process had to do with my the issue I was facing, as the error messages didn't reflect what was actually wrong.

PR #649, opened 2025-10-27
Added multipart imagery upload capability

FeezyHendrix:

- [x] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation
- [ ] 🧑‍💻 Refactor
- [ ] ✅ Test
- [x] 🤖 Build or CI

## Related Issue

Implements resumable multipart uploads for drone imagery

## Describe this PR

This PR implements a comprehensive drone image upload and processing workflow with the following key features:

**Frontend:**
- Integrated **Uppy** file uploader with AWS S3 multipart upload support for reliable, resumable uploads
- Implemented custom themed UI matching the app's design system (red #D73F3F, Manrope font)
- Created multi-step drone image processing workflow with 4 stages: Upload → Classification → Review → Processing
- Added real-time upload progress tracking and queue management
- Removed single-file upload in favor of multipart upload for better performance with large imagery files

**Backend:**
- Implemented complete S3 multipart upload API endpoints 
- Added **project_images** database table 
- Integrated **EXIF data extraction** using Pillow 
- Implemented **ARQ background task processing** for async image processing to avoid blocking API responses
- Added **idempotent duplicate detection** using MD5 hash to prevent duplicate uploads from creating multiple database records
- Created image status tracking (staged, classified, invalid_exif, unmatched, duplicate)

**Infrastructure:**
- Fixed Docker networking issues with MinIO by adding `extra_hosts` configuration for localhost resolution
- Reverted from boto3 back to MinIO SDK for better presigned URL signature handling
- Added Redis service for ARQ task queue
- Added Pillow dependency for EXIF extraction

**Technical Highlights:**
- Presigned URLs are generated with correct hostname to prevent signature mismatches
- Background processing ensures fast API responses while handling expensive EXIF extraction asynchronously
- Duplicate handling is idempotent: re-uploading the same file returns existing record without creating duplicates
- PostGIS POINT type used for spatial queries on image locations
- Comprehensive indexing on project_id, task_id, status, hash_md5, and spatial location

## Screenshots

The Uppy file uploader with custom theming and multipart upload progress:

WorkFlow Button 
Screenshot 2025-10-27 at 13 45 51

Initial View
Screenshot 2025-10-27 at 13 46 48

Pre Upload View
Screenshot 2025-10-27 at 13 47 09

Post Upload View
Screenshot 2025-10-27 at 13 47 17

### Testing the Upload Flow:

1. **Start services:**
   ```bash
   docker compose up -d
Navigate to a project and open the drone image upload section
Test multipart upload:
Select multiple large drone images (>5MB recommended)
Verify upload progress is displayed
Check uploads complete successfully
Verify images appear in the staging area
Test duplicate handling:
Upload the same image twice
Verify no duplicate database records are created
Check that the second upload returns existing record

Verify background processing:
Check ARQ worker logs
docker compose logs -f arq-worker

Verify images in database with EXIF data
Check GPS coordinate extraction:
Upload images with GPS EXIF data
Verify location field is populated in database
Verify exif JSONB contains camera settings

Key Files to Review:
Backend:
[src/backend/app/s3.py](vscode-webview://1uop7q036khfrhrkep55v12l1028j40ul96hsbislfmga3h1j2cl/src/backend/app/s3.py) - S3 multipart upload implementation
[src/backend/app/projects/project_routes.py](vscode-webview://1uop7q036khfrhrkep55v12l1028j40ul96hsbislfmga3h1j2cl/src/backend/app/projects/project_routes.py) - Upload API endpoints
[src/backend/app/images/image_logic.py](vscode-webview://1uop7q036khfrhrkep55v12l1028j40ul96hsbislfmga3h1j2cl/src/backend/app/images/image_logic.py) - EXIF extraction logic
[src/backend/app/arq/tasks.py](vscode-webview://1uop7q036khfrhrkep55v12l1028j40ul96hsbislfmga3h1j2cl/src/backend/app/arq/tasks.py) - Background image processing task
[src/backend/app/migrations/versions/001_add_project_images_table.py](vscode-webview://1uop7q036khfrhrkep55v12l1028j40ul96hsbislfmga3h1j2cl/src/backend/app/migrations/versions/001_add_project_images_table.py) - Database schema
Frontend:
[src/frontend/src/components/DroneOperatorTask/DescriptionSection/UppyFileUploader/](vscode-webview://1uop7q036khfrhrkep55v12l1028j40ul96hsbislfmga3h1j2cl/src/frontend/src/components/DroneOperatorTask/DescriptionSection/UppyFileUploader/) - Uppy integration
[src/frontend/src/components/DroneOperatorTask/DescriptionSection/DroneImageProcessingWorkflow/](vscode-webview://1uop7q036khfrhrkep55v12l1028j40ul96hsbislfmga3h1j2cl/src/frontend/src/components/DroneOperatorTask/DescriptionSection/DroneImageProcessingWorkflow/) - 4-step workflow
Infrastructure:
[compose.yaml](vscode-webview://1uop7q036khfrhrkep55v12l1028j40ul96hsbislfmga3h1j2cl/compose.yaml) added extra_hosts configuration

Checklist before requesting a review
 📖 Read the HOT Code of Conduct: https://docs.hotosm.org/code-of-conduct
 👷‍♀️ Create small PRs. In most cases, this will be possible.
 ✅ Provide tests for your changes.
 📝 Use descriptive commit messages.
 📗 Update any related documentation and include any relevant screenshots.
[optional] What gif best describes this PR or how it makes you feel?
Drone taking off smoothly Smooth takeoff with resumable uploads! ✈️

### DIFF
--- a/src/backend/app/images/image_logic.py
+++ b/src/backend/app/images/image_logic.py
@@ -24,14 +24,18 @@ def _convert_exif_value(value: Any) -> Any:
     which need to be converted to standard Python types for JSON serialization.
     """
     # Handle IFDRational (PIL's rational number type)
-    if hasattr(value, 'numerator') and hasattr(value, 'denominator'):
+    if hasattr(value, "numerator") and hasattr(value, "denominator"):
         # Convert rational to float
-        return float(value.numerator) / float(value.denominator) if value.denominator != 0 else 0.0
+        return (
+            float(value.numerator) / float(value.denominator)
+            if value.denominator != 0
+            else 0.0
+        )
 
     # Handle bytes
     if isinstance(value, bytes):
         try:
-            return value.decode('utf-8', errors='ignore')
+            return value.decode("utf-8", errors="ignore")
         except:
             return str(value)
 
@@ -258,7 +262,7 @@ async def check_duplicate_image(
     Returns:
         UUID of the duplicate image if found, None otherwise
     """
-    sql = f"""
+    sql = """
         SELECT id FROM project_images
         WHERE project_id = %(project_id)s
         AND hash_md5 = %(hash_md5)s
@@ -282,7 +286,7 @@ async def mark_image_as_duplicate(
         image_id: ID of the image to mark as duplicate
         duplicate_of: ID of the original image
     """
-    sql = f"""
+    sql = """
         UPDATE project_images
         SET status = %(status)s, duplicate_of = %(duplicate_of)s
         WHERE id = %(image_id)s
--- a/src/backend/scripts/migrate_images_from_s3.py
+++ b/src/backend/scripts/migrate_images_from_s3.py
@@ -19,7 +19,9 @@ from app.db.db_models import DbProject, DbTask
 from app.models.enums import ImageStatus
 
 
-S3_PUBLIC_ENDPOINT = os.getenv("S3_PUBLIC_ENDPOINT", "https://drone-tm-public.s3.amazonaws.com")
+S3_PUBLIC_ENDPOINT = os.getenv(
+    "S3_PUBLIC_ENDPOINT", "https://drone-tm-public.s3.amazonaws.com"
+)
 
 
 def fetch_images_json(project_id: str, task_id: str) -> list | None:
@@ -31,7 +33,9 @@ def fetch_images_json(project_id: str, task_id: str) -> list | None:
             return json.loads(data)
     except HTTPError as e:
         if e.code == 404:
-            log.warning(f"images.json not found for project {project_id}, task {task_id}")
+            log.warning(
+                f"images.json not found for project {project_id}, task {task_id}"
+            )
         else:
             log.error(f"Failed to fetch {url}: {e.code}")
         return None
@@ -49,7 +53,7 @@ def migrate_images_for_task(
     session: Session,
     project_id: uuid.UUID,
     task_id: uuid.UUID,
-    uploaded_by: str | None = None
+    uploaded_by: str | None = None,
 ) -> int:
     images_data = fetch_images_json(str(project_id), str(task_id))
 
@@ -130,7 +134,7 @@ def migrate_images_for_task(
                 "uploaded_by": uploaded_by,
                 "uploaded_at": datetime.now(timezone.utc),
                 "status": ImageStatus.UPLOADED,
-            }
+            },
         )
 
         upserted_count += 1
@@ -139,10 +143,7 @@ def migrate_images_for_task(
 
 
 def migrate_all_projects():
-    engine = create_engine(
-        settings.DTM_DB_URL.unicode_string(),
-        echo=False
-    )
+    engine = create_engine(settings.DTM_DB_URL.unicode_string(), echo=False)
 
     with Session(engine) as session:
         projects = session.query(DbProject).all()
@@ -154,18 +155,13 @@ def migrate_all_projects():
         for project in projects:
             log.info(f"Processing project: {project.name} ({project.id})")
 
-            tasks = session.query(DbTask).filter(
-                DbTask.project_id == project.id
-            ).all()
+            tasks = session.query(DbTask).filter(DbTask.project_id == project.id).all()
 
             log.info(f"  Found {len(tasks)} tasks for project {project.id}")
 
             for task in tasks:
                 count = migrate_images_for_task(
-                    session,
-                    project.id,
-                    task.id,
-                    uploaded_by=project.author_id
+                    session, project.id, task.id, uploaded_by=project.author_id
                 )
 
                 if count > 0:
--- a/src/backend/app/arq/tasks.py
+++ b/src/backend/app/arq/tasks.py
@@ -1,5 +1,4 @@
 import asyncio
-import json
 from typing import Any, Dict, Optional
 from uuid import UUID
 
@@ -178,7 +177,7 @@ async def process_uploaded_image(
             duplicate_id = await check_duplicate_image(db, UUID(project_id), file_hash)
             if duplicate_id:
                 log.info(f"Duplicate detected: {file_hash} -> {duplicate_id}")
-                sql = f"SELECT * FROM project_images WHERE id = %(id)s"
+                sql = "SELECT * FROM project_images WHERE id = %(id)s"
                 async with db.cursor(row_factory=dict_row) as cur:
                     await cur.execute(sql, {"id": str(duplicate_id)})
                     existing_record = await cur.fetchone()
@@ -200,7 +199,9 @@ async def process_uploaded_image(
                 exif_dict, location = extract_exif_data(file_content)
 
                 if exif_dict:
-                    log.info(f"✓ EXIF: {len(exif_dict)} tags | GPS: {location is not None}")
+                    log.info(
+                        f"✓ EXIF: {len(exif_dict)} tags | GPS: {location is not None}"
+                    )
                     log.debug(f"EXIF tags: {list(exif_dict.keys())[:10]}")
                 else:
                     log.warning(f"✗ No EXIF data in: {filename}")
--- a/src/backend/app/db/db_models.py
+++ b/src/backend/app/db/db_models.py
@@ -243,11 +243,19 @@ class DbProjectImage(Base):
     id = cast(str, Column(UUID(as_uuid=True), primary_key=True))
     project_id = cast(
         str,
-        Column(UUID(as_uuid=True), ForeignKey("projects.id", ondelete="CASCADE"), nullable=False),
+        Column(
+            UUID(as_uuid=True),
+            ForeignKey("projects.id", ondelete="CASCADE"),
+            nullable=False,
+        ),
     )
     task_id = cast(
         str,
-        Column(UUID(as_uuid=True), ForeignKey("tasks.id", ondelete="SET NULL"), nullable=True),
+        Column(
+            UUID(as_uuid=True),
+            ForeignKey("tasks.id", ondelete="SET NULL"),
+            nullable=True,
+        ),
     )
     filename = cast(str, Column(Text, nullable=False))
     s3_key = cast(str, Column(Text, nullable=False))
@@ -275,7 +283,9 @@ class DbProjectImage(Base):
     # Relationships
     project = relationship(DbProject, backref="images")
     task = relationship(DbTask, backref="images")
-    uploader = relationship(DbUser, foreign_keys=[uploaded_by], backref="uploaded_images")
+    uploader = relationship(
+        DbUser, foreign_keys=[uploaded_by], backref="uploaded_images"
+    )
 
     __table_args__ = (
         Index("idx_project_images_project_id", "project_id"),
--- a/src/backend/app/gcp/gcp_crud.py
+++ b/src/backend/app/gcp/gcp_crud.py
@@ -345,7 +345,9 @@ async def find_images_in_a_task_for_point(
     s3_images_json_path = f"dtm-data/projects/{project_id}/{task_id}/images.json"
 
     # Generate pre-signed URL for the `images.json` file
-    s3_images_json_url = generate_presigned_download_url(settings.S3_BUCKET_NAME, s3_images_json_path)
+    s3_images_json_url = generate_presigned_download_url(
+        settings.S3_BUCKET_NAME, s3_images_json_path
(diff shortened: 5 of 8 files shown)


compose.yaml
```diff
@@ -28,6 +28,8 @@ services:
       - ${PROJECT_DIR:-.}/src/backend/packages/drone-flightplan/drone_flightplan:/opt/python/lib/python3.11/site-packages/drone_flightplan:ro
       - frontend-html:/project/src/backend/frontend_html
     env_file: .env
+    extra_hosts:
```
spwoodcock, line 31: What was the networking issue you were having that requires this? Everything should be running via internal docker networking / service names, instead of localhost

src/backend/app/projects/project_routes.py
```diff
@@ -332,39 +338,53 @@ async def generate_presigned_url(
...
-                        client.list_objects(
-                            settings.S3_BUCKET_NAME, image_dir, recursive=True
-                        ),
-                    )
-
-                    # Remove the objects (images)
-                    errors = client.remove_objects(
-                        settings.S3_BUCKET_NAME, delete_object_list
+                    # List objects to delete
+                    paginator = client.get_paginator("list_objects_v2")
```
spwoodcock, line 342: This process made sense uploading for a specific task, but probably not for an entire project.

One user could click to 'replace existing' imagery, deleting all of the imagery collected by all users on all tasks in the project (potentially undoing many days of fieldwork).

Perhaps this code can be reused on an admin-only endpoint to clear all the existing imagery and reset the project? Or probably better / safer would be for the admin to reset a single task only, deleting the imagery related to that task.

spwoodcock: Also something like this definitely shouldn't be running the the 'get-presigned-url' endpoint. Consider this endpoint could be called thousands of times in a given upload period. There should be as minimal logic as possible here ideally 👍

spwoodcock: Actually, I was confused by the function name and docstring here. Based on the return signature, it actually returning a list of pre-signed URLs, so this will be called less than I was anticipating. But still, the variable names are confusing (I know you are building on what was already there / badly named already, so probably not your fault!). Would be great if we could clarify the variable / function / docstring var names to make it easier to read 😄

src/backend/app/projects/project_routes.py
```diff
@@ -332,39 +338,53 @@ async def generate_presigned_url(
...
                     )
 
             # Generate a new pre-signed URL for the image upload
-            url = client.get_presigned_url(
-                "PUT",
-                settings.S3_BUCKET_NAME,
-                image_path,
-                expires=timedelta(hours=data.expiry),
+            # Use public endpoint for presigned URLs in local dev
+            presigned_client = s3_client(use_public_endpoint=True)
```
spwoodcock: Having a variable `use_public_endpoint` to account for local development isn't terrible, but it does muddy local development / production usage. It would be much nicer if there were a pattern to intercept the URLs and update them if it's local development, or something similar. I wonder if there is a cleaner approach that can live in s3.py, instead of being spread throughout the codebase

src/backend/app/projects/project_routes.py
spwoodcock: Scrolling down, I now see `/sign-part-upload`, which I assume is the target endpoint for generating pre-signed URLs via uppy. I wonder if we even need `/generate-presigned-url` for anything now? Just speculating

src/backend/app/projects/project_routes.py
```diff
@@ -746,6 +766,176 @@ async def upload_imagery_to_oam(
     return {"message": "Uploading to OAM Started", "status": OAMUploadStatus.UPLOADING}
 
 
+@router.post("/initiate-multipart-upload/", tags=["Image Upload"])
```
spwoodcock, line 768: These endpoints all look pretty good. Minimal, whats needed for S3 spec / uppy 👍

src/backend/app/utils.py
```diff
@@ -593,7 +593,7 @@ def calculate_flight_time_from_placemarks(placemarks: Dict) -> Dict:
     }
 
 
-def strip_presigned_url_for_local_dev(url: str) -> str:
+def strip_presigned_url_for_local_dev(url: str, strip_presign: bool = True) -> str:
```
spwoodcock, line 602: This works, but now we have logic in two places for handling local dev / public S3 endpoints. Would be nice to consolidate (as mentioned in previous comment)

src/backend/pyproject.toml
```diff
@@ -20,7 +20,7 @@ dependencies = [
     "requests>=2.32.3",
     "requests-oauthlib>=2.0.0",
     "loguru>=0.7.2",
-    "minio>=7.2.7",
+    "minio>=7.2.0",
```
spwoodcock: This won't fix it, as `>=` was specified. You will need to use a hard pin `==` to ensure the version is the one you expect

src/frontend/src/components/DroneOperatorTask/DescriptionSection/DroneImageProcessingWorkflow/index.tsx
```diff
@@ -0,0 +1,169 @@
...
+}
+
+const DroneImageProcessingWorkflow = ({
+  isOpen,
+  onClose,
+  projectId,
+}: IDroneImageProcessingWorkflowProps) => {
+  const [currentStep, setCurrentStep] = useState(1);
+
+  const steps = [
```
spwoodcock, line 23: This is good - did you see the existing `StepSwitcher` component used during project creation though? Would it make sense to reuse this, as it's basically the same sort of workflow?

src/frontend/src/components/DroneOperatorTask/DescriptionSection/UppyFileUploader/index.tsx
```diff
@@ -0,0 +1,272 @@
...
+        );
+
+        if (imageFiles.length > 0) {
+          const exifDataPromises = imageFiles.map(async file => {
+            const exifData = await getExifData(file.data);
+            return exifData;
+          });
+
+          const exifData = await Promise.all(exifDataPromises);
+          dispatch(setFilesExifData(exifData));
```
spwoodcock, line 214: Assume you're having to do this for lack of imagery with correct EXIF data?

compose.yaml
FeezyHendrix: The issue is with signing the part upload url. If we sign with the internal url http://minio and replace it with http://localhost:9000, we get an error, because minIO is expecting `http://minio`.

src/backend/app/projects/project_routes.py
FeezyHendrix: Haha no worries, I would take a look into the functions and try refactor them to be more explicit

src/backend/app/projects/project_routes.py
FeezyHendrix: We can introduce a APP_ENV variable to control how the behavior is expected on different environments. This would ensure the checks and configuration happen inside s3.py

src/frontend/src/components/DroneOperatorTask/DescriptionSection/DroneImageProcessingWorkflow/index.tsx
FeezyHendrix: Oh, I missed it, would refactor to use the component

src/frontend/src/components/DroneOperatorTask/DescriptionSection/UppyFileUploader/index.tsx
FeezyHendrix: Yes, I am.

compose.yaml
spwoodcock: Right, nice, not a bad approach then!

This way we can use http://localhost:9000 for all minio functions on local dev, right? Upload, signing, download, etc

src/backend/app/projects/project_routes.py
spwoodcock: Not a bad idea, but we don't use that pattern anywhere else & it seems a bit cumbersome.

Could we determine this automatically, if S3_ENDPOINT is localhost or the docker service name?

src/backend/pyproject.toml
spwoodcock: That said, whenever these pins were specified as `>=` that was done incorrectly - all pins in **applications** should be `==`, as per https://docs.hotosm.org/dev-guide/repo-management/dependencies/#locking-dependency-versions.

If you get a chance to update these while changing the `minio` pin, please do! 🙏

src/backend/app/images/image_logic.py
```diff
@@ -0,0 +1,284 @@
...
+    """
+    # Convert location dict to PostGIS point if provided
+    location_sql = None
+    if image_data.location:
+        lat = image_data.location.get("lat")
+        lon = image_data.location.get("lon")
+        if lat is not None and lon is not None:
+            location_sql = f"ST_SetSRID(ST_MakePoint({lon}, {lat}), 4326)"
+
+    sql = """
```
spwoodcock: f-strings are generally more readable, but this works 👍

src/backend/app/projects/project_routes.py
```diff
@@ -746,6 +766,193 @@ async def upload_imagery_to_oam(
...
+        # Complete the multipart upload in S3
+        complete_multipart_upload(
+            settings.S3_BUCKET_NAME,
+            data.file_key,
+            data.upload_id,
+            data.parts,
+        )
+
+        # Queue background task to process image (EXIF extraction, hash, duplicate check)
+        job = await redis.enqueue_job(
```
spwoodcock, line 872: We will queue a job for each image, meaning thousands of jobs being queued.

I was thinking in prod it may be good to batch this, but on second thought, it would potentially lead to issues where it's not easy to determine 'bad' images. If a single image fails e.g. on EXIF extraction, then the whole batch would fail.

I think redis / arq should be able to handle processing thousands of jobs, but we should keep an eye on the performance of this & if we need to tune Redis to be able to handle everything.

Perhaps a comment in the code along these lines would help out a future dev debugging 😄 
(comments in general are very useful!)

src/backend/app/migrations/versions/001_add_project_images_table.py
spwoodcock, line 1: It appears we have a migration here, but no changes to the db_models.py file?

Did you generate the migration, but forget to push the changes to the models?

compose.yaml
FeezyHendrix: Yes, we can just use localhost:9000, for everything.

FeezyHendrix:
How are you thinking of the migration? Essentially, a simple python script?

src/backend/app/projects/project_routes.py
FeezyHendrix: Well, one way to do this is to probably check if it's localhost itself, by checking the string?

spwoodcock: Yep could work!

`if localhost in string` or `string is http://s3:9000`

src/backend/app/projects/project_routes.py
FeezyHendrix: Added the comment for this

src/backend/scripts/migrate_images_from_s3.py
```diff
@@ -0,0 +1,166 @@
...
+from sqlalchemy.orm import Session
+
+sys.path.insert(0, str(Path(__file__).parent.parent))
+
+from app.config import settings
+from app.db.db_models import DbProject, DbProjectImage, DbTask
+from app.models.enums import ImageStatus
+
+
+S3_PUBLIC_ENDPOINT = os.getenv("S3_PUBLIC_ENDPOINT", "https://drone-tm-public.s3.amazonaws.com")
```
spwoodcock: Leaving a note here to remember to switch this to a pre-signed download from a private bucket when we roll this out to prod

src/backend/scripts/migrate_images_from_s3.py
spwoodcock, line 1: This looks good!

The only comment I would make is that is pretty good practice to make migrations idempotent, so they can be run multiple times without consequence (e.g. using upserts instead of inserts).

What would happen if the script fails part way & we need to determine which entries correctly migrated, and which ones didn't?

PR #657, opened 2025-11-04
Setup fixes

stanleypliu:

- [ ] 🍕 Feature
- [x] 🐛 Bug Fix
- [ ] 📝 Documentation
- [ ] 🧑‍💻 Refactor
- [ ] ✅ Test
- [x] 🤖 Build or CI

## Related Issue

N/A

## Describe this PR

~~I noticed a couple of small issues during the setup of this repository~~
~~- the migrations Docker service was starting up prior to the database and backend services which led to a race condition; I edited it to be dependent on those services.~~
~~- `envsubst` doesn't take the `-i` flag nor the `-o` flag as far as I can tell. I could be wrong about this but I checked its arguments and they don't seem to exist, hence I replaced its usage in `setup.md` with `cp`.~~

EDIT: Was incorrectly setting up the repository using `compose.sub.yaml` - in the end, I just edited the `setup.md` file to reflect use of the Justfile when setting up 

## Screenshots

Please provide screenshots of the change.

N/A

## Review Guide

If able, perform a clean setup from scratch with the old steps and compare against the new changes.

### DIFF
--- a/docs/dev/setup.md
+++ b/docs/dev/setup.md
@@ -2,8 +2,15 @@
 
 ## Prerequisites
 
-- [Docker](https://docs.docker.com/engine/install)
-- [Docker Compose](https://docs.docker.com/compose/install)
+!!! important
+
+    You must have Docker and Docker Compose installed first:
+
+    - [Docker install guide](https://docs.docker.com/engine/install)
+    - [Docker Compose install guide](https://docs.docker.com/compose/install)
+
+Just is required too:
+
 - [Just](https://github.com/casey/just#installation)
 
 ## (Optional) Configure Google OAuth
@@ -34,11 +41,7 @@
 - `compose.sub.yaml` is used for production builds - `compose.yaml`, meanwhile, is used for development.
 - To start the application, run:
   `just start all`
-- This single command will:
-  1.  Generate a `.env` file if one doesn't exist.
-  2.  Pull the required Docker images.
-  3.  Build the frontend container.
-  4.  Start all services correctly.
+- This command will pull the required Docker images and start all services correctly.
 
 ### Access The Services
 


compose.sub.yaml
```diff
@@ -145,8 +146,10 @@ services:
   migrations:
     image: ghcr.io/hotosm/drone-tm/backend:${GIT_BRANCH:-main}
     depends_on:
-      - backend
-      - db
+      backend:
```
spwoodcock: Actually the migrations should run before the backend I believe, so the conditions ordering should be reversed: backend depends on migrations

spwoodcock: Its a bit strange this was set up like it was

docs/dev/setup.md
```diff
@@ -20,7 +20,7 @@
 ## Configure Your Dotenv
 
 - There is an example `.env.example` file that `.env` can be generated from:
-  `envsubst -i .env.example -o .env`
```
spwoodcock, line 23: Its using a8m/envsubst, so that should be clearer, sorry. IMO all the setup should be included in the Justfile, avoiding the need for the user to install dependencies etc

spwoodcock (commented):
Thanks for this! Its a good prompt that we dont really have any documentation and the config is currently not great 🙏

Just checking, are you using compose.sub.yaml for local dev, or you're just debugging prod?

stanleypliu:
> Thanks for this! Its a good prompt that we dont really have any documentation and the config is currently not great 🙏
> 
> Just checking, are you using compose.sub.yaml for local dev, or you're just debugging prod?

I'm using it locally, was following the instructions in `setup.md`. I didn't know that there was a `Justfile` that took care of the setup - I'll update the document about that.

spwoodcock (approved):

docs/dev/setup.md
```diff
@@ -1,12 +1,10 @@
 # Setting Up DroneTM For Development
 
-!!! important
```
spwoodcock, line 3: I noticed the link in the docs site was broken to the dev setup page:

https://docs.dronetm.hotosm.org/dev/setup/
(I just fixed the link so it opens from the sidebar)

These adminition blocks are useful for a nicer display in mkdocs - could we revert to keep this part in an admonition block please 🙏 
image

docs/dev/setup.md
```diff
@@ -19,8 +17,8 @@
 
 ## Configure Your Dotenv
 
-- There is an example `.env.example` file that `.env` can be generated from:
-  `envsubst -i .env.example -o .env`
+- There is an example `.env.example` file that `.env` can be generated from using this command:
+  `just config generate-dotenv`
```
spwoodcock, line 28: Perfect, thanks for adding this!

docs/dev/setup.md
```diff
@@ -30,19 +28,17 @@
...
-
-```bash
-docker compose -f compose.yaml -f compose.odm.yaml up -d
-```
+### Build and Run The Containers
+
+- The `Justfile` in the project root contains all the logic needed to build and run the application.
+- `compose.sub.yaml` is used for production builds - `compose.yaml`, meanwhile, is used for development.
+- To start the application, run:
+  `just start all`
```
spwoodcock, line 43: This part is kind of correct.

I think you may have confused the `just start prod` command with `just start all`?
`just start prod` will do all you mention below, but `just start all` hasn't been configured to do this.
But it would be nice change to make it do that!

stanleypliu: Good point! I've removed the steps it doesn't cover

spwoodcock (approved):

Issue #641, opened 2025-10-16
Cleanup NodeODM job after processing is complete


spwoodcock:
- It would be really nice if DroneTM did its own cleanup pushing a delete command via API on NodeODM.
- This way the timeout on jobs can be set to be quite high, knowing most of them will be automatically deleted once done.
- Also any processing with issues can be automatically identified in NodeODM and restarted or cleaned up as needed.

PR #658, opened 2025-11-05
Implementing cleanup of NodeODM tasks

allisonsibrian:

- [X] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation
- [ ] 🧑‍💻 Refactor
- [ ] ✅ Test
- [ ] 🤖 Build or CI

## Related Issue

Adds: Cleanup NodeODM job after processing is complete #641

## Describe this PR

This PR implements the automatic cleanup on the NodeODM server after the processing job is completed or fails.

I added a call to task.remove() within the finally block of the process_assets_from_odm function for the cleanup process.

I attempted different ways to test the changes locally doing end-to-end testing, since it's easier to track logs, but encountered some environment configuration issues:
- When using the local UI to test: the front-end login wasn't functional due to the dummy credentials.
- Using the NodeODM Web UI for callback: the callback request from the NodeODM container was blocked with a `403 Forbidden` error, which prevented my log messages from appearing in the backend logs to actually check if cleanup process occurred.

I'm unable to confirm changes because of these issues, and am just creating this as a draft PR to see if the CI tests work or to solicit advice on addressing the test issues above. 

## Review Guide

I feel like end-to-end testing seems like the right approach to confirm these changes (let me know if I am mistaken or if an alternate testing method is better in this case):
- Trigger a NodeODM processing job.
- After the job finishes, monitor the backend logs for confirmation messages that were added:
       - Attempting to delete task {odm_task_id} from NodeODM.
       - Successful attempt at deleting task {odm_task_id} from NodeODM.
-Verify the task no longer appears in the NodeODM UI list.

### DIFF
--- a/src/backend/app/projects/image_processing.py
+++ b/src/backend/app/projects/image_processing.py
@@ -524,9 +524,13 @@ async def process_assets_from_odm(
             try:
                 log.info(f"Attempting to delete task {odm_task_id} from NodeODM.")
                 task.remove()
-                log.info(f"Successful attempt at deleting task {odm_task_id} from NodeODM.")
-            except Exception as e: 
-                log.error(f"Error occurred while cleaning up task {odm_task_id} from NodeODM: {e}.")
+                log.info(
+                    f"Successful attempt at deleting task {odm_task_id} from NodeODM."
+                )
+            except Exception as e:
+                log.error(
+                    f"Error occurred while cleaning up task {odm_task_id} from NodeODM: {e}."
+                )
 
         if os.path.exists(output_file_path):
             try:


spwoodcock (approved):
Looks like a pretty reasonable approach to me!

Two ways of testing:
1 Connect an ODM instance and run the processing manually. View the log outputs etc to confirm.
2. e2e tests would be excellent, as you mention, but I'm not sure we have this all in place, and is probably outside the scope of this PR. If you have time to dabble with an e2e test setup (perhaps using playwright), then feel free to try! 🙏

allisonsibrian:
> To me, I feel like the code logic makes sense, but it's just the testing configuration preventing seeing the log outputs. So I'm not sure if I can mark this as ready for review or if you advise to keep troubleshooting to confirm the log messages to be 100% sure it works.

Just wanted to bump this message

spwoodcock:
Apologies - I thought I would have time to test this myself, but under too much time pressure.

We need to test locally, as a big change was just merged to dev that we probably can't afford to mix with this right now.

The key is that there are no NodeODM instances attached to the default compose config.

If you wish to test the processing, I would suggest adding a nodeodm container as part of compose.test.yaml and then updating your env to use the compose service name

`http://nodeodm:9000` (I think port 9000 is the default from memory)

spwoodcock:
Thank you @allisonsibrian !!

spwoodcock:
Looks great! 🙌

In the logs, between the triggering of processing and downloading of results there is only a few miliseconds.

Just checking, did any processing actually trigger? Normally ODM take many minutes, up to an hour to complete full processing, as it has many stages. Assuming you might be mocking the result somewhere?

allisonsibrian:
Yea, the full processing didn't occur since the job failed at the image download stage. I encountered S3 connectivity issues (403 Forbidden/404 Not Found errors) between the backend worker and the local MinIO container, which exited the process immediately.

I still wanted to see if the cleanup happened, and the logs confirm that even though the job crashed hard during the setup, the finally block is still triggered and removes the task from NodeODM.

I can continue debugging the S3 authentication/network configurations to ensure that the full process is happening and removing the task from NodeODM.

I can add the compose.test.yaml and the final verification logs once I resolve the environment configuration!

spwoodcock:
Oh makes sense, thanks!

I guess the test checks the new cleanup functionality works as intended then, so we could merge this, then continue with the dev testing setup in another PR 👍

src/backend/uv.lock
spwoodcock, line 1: Sorry to be a pain - any idea why this caused a lockfile change?

allisonsibrian: Sorry this was accidental, I think it was in the very beginning when I was trying to set up the local testing environment, but I can revert this file

spwoodcock (approved):

Issue #654, opened 2025-11-01
Redirects to specific frontend URLs do not work if the user is not logged in


spwoodcock:
- Logged in, the user can go straight to a project via direct link.
- Not logged in, they go the home page, and aren't prompted to be logged in.
- We need to fix this to open the prompt for the user to log in, then direct them back to the page they should have entered.

emi420:
Maybe this one can be also fixed:

https://github.com/hotosm/drone-tm/issues/550

Basically it should be possible to open new tabs from any link in DroneTM, something that will improve accessibility by following standards and expected browser behaviors.

stanleypliu:
I'm happy to take this on as my first issue. I would like to know more about what routes we'd want to permit, since you mentioned the word "specific" in the title. Are there any routes where we'd just want to redirect the user to /projects?

spwoodcock:
Perhaps thats ambiguously worded: by specific I was mostly meaning any subpath other than `/`

So `/projects/hsidj8e6e8dhsksjsjdb`
Or `/login`

If you go directly to these URLs, it doesn't work. If you access the frontend from `/`, then route using the SPA, it works 😃

A key URL people need to address is directly linking to a project, so it can be shared amongst operators

stanleypliu:
I tested this on production and it works for both project creators and drone operators. OK to close this?

spwoodcock:
Yes definitely! Sorry for missing this 😄

PR #659, opened 2025-11-05
Implement redirects to pages behind the login

stanleypliu:

- [x] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation
- [ ] 🧑‍💻 Refactor
- [ ] ✅ Test
- [ ] 🤖 Build or CI

## Related Issue

Addresses #654 

## Describe this PR

Aims to address logged-out users from not being able to access protected routes (i.e. behind the login wall). This should work for both Google users and standard users. This relies on setting a property on `sessionStorage` and utilising that to determine whether or not to show the sign-in overlay and toast when the user gets redirected back to the homepage. It also addresses open redirects by disallowing them if they are not a relative path.

## Screenshots

https://github.com/user-attachments/assets/aef952fd-6915-4b44-9a08-6279c6d22c22

https://github.com/user-attachments/assets/fdddd9d0-667e-4738-afbd-5979b9d28c3c

For addressing the open redirects, I was looking at having a whitelist of pages as the most secure way of redirecting the user, but since I'm not cognizant of all the different routes in this platform (and with the difficulty of maintaining such a list) I went with checking that the root domain is still that of the site. 

## Review Guide

Flow is as follows:
- User attempts to access a route, e.g. /projects/123
- User is redirected back to the homepage, with the sign-in overlay open and a toast instructing them to sign in.
- User authenticates via either route.
- User lands on the specific route, instead of /projects or /complete-profile.
- User clicks sign out, and when signing in again, there shouldn't be any toast (implying that `location.state.from`, which is used to determine whether or not to show the overlay, has been cleared).

### DIFF
--- a/src/frontend/src/modules/user-auth-module/src/components/Authentication/Login/index.tsx
+++ b/src/frontend/src/modules/user-auth-module/src/components/Authentication/Login/index.tsx
@@ -4,6 +4,7 @@ import { useState, useEffect } from 'react';
 import { useLocation, useNavigate } from 'react-router-dom';
 import { useForm } from 'react-hook-form';
 import { useQuery, useMutation } from '@tanstack/react-query';
+import { toast } from 'react-toastify';
 
 import Image from '@Components/RadixComponents/Image';
 import { Input, Label, FormControl } from '@Components/common/FormUI';
@@ -15,7 +16,6 @@ import { useTypedDispatch } from '@Store/hooks';
 import { signInGoogle, signInUser } from '@Services/common';
 import { setUserState } from '@UserModule/store/actions/user';
 import googleIcon from '@Assets/images/google-icon.svg';
-import { toast } from 'react-toastify';
 import { isSafeRedirect } from '@Utils/url';
 
 const { BASE_URL } = process.env;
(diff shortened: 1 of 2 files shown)


stanleypliu:
> This solves the problem of redirect after login.
> 
> In prod, we have another issue that the bunkerweb proxy gives a 404 for URLs other than the home page. Do you think this might fix that? (we can test on dev.dronetm.org)

I'm not sure it would be related (but I don't know how it is working either), we'd have to inspect the logs to see what the proxy is doing. I presume dev.dronetm.org has that proxy too?

spwoodcock:
> > This solves the problem of redirect after login.
> > 
> > In prod, we have another issue that the bunkerweb proxy gives a 404 for URLs other than the home page. Do you think this might fix that? (we can test on dev.dronetm.org)
> 
> I'm not sure it would be related (but I don't know how it is working either), we'd have to inspect the logs to see what the proxy is doing. I presume dev.dronetm.org has that proxy too? 
> 

Yep it does!

Needs some testing, likely of the request / response headers.

I feel we might be able to catch the request and redirect to the login page if needed

PR #664, opened 2025-11-14
Suggest task split to user if flight time exceeds 80% of battery life 

### LINKED ISSUE #607: Suggest to user that a task should be split if the flight time is too high
- We calculate approximate flight time.
- If they are using a drone that has it's battery time saved at a DroneTM stat, we should use that!
- Obviously the user doesn't want to use a DJI Mini 4 Pro battery of ~20 minute flyable time, when the task area is 22 minutes.

stanleypliu:

- [x] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation
- [ ] 🧑‍💻 Refactor
- [ ] ✅ Test
- [ ] 🤖 Build or CI

## Related Issue

Fixes #607 

## Describe this PR

- Provides battery life estimates for the DJI drones and the Potensic Atom 2 within `drone-flightplan`.
- Updates calculations to provide the battery life warning to the frontend.
- Displays the warning to the user if the criteria is satisfied, as well as showing a disclaimer about the calculated flight time. 

## Screenshots

Screenshot 2025-11-15 at 14 27 14

I was considering creating actual drone records in the Drones table within the drone-tm database, which also has a battery life column, but since the drone-flightplan package is used by other repositories, it made more sense to have this information commonly available. It also appears that these tables are unused.

## Review Guide

Log in as either a drone operator/project creator. Look for projects with flight times that exceed 25 minutes on a particular task. Go to the task. The toast should show (and reappear every time a different drone type is selected that triggers the condition).
Alternatively, as a project creator, create a project with sufficiently large task areas (use task lengths greater than 600 m). After creating, access the project and go to a particular task.

### DIFF
--- a/src/backend/packages/drone-flightplan/drone_flightplan/calculate_parameters.py
+++ b/src/backend/packages/drone-flightplan/drone_flightplan/calculate_parameters.py
@@ -58,7 +58,9 @@ def _calculate_constants(
     print(f"GSD To AGL Const:     {gsd_to_agl_const}")
 
 
-def calculate_adjusted_max_battery_life(drone_type: DroneType, ground_speed: float) -> float:
+def calculate_adjusted_max_battery_life(
+    drone_type: DroneType, ground_speed: float
+) -> float:
     """
     Calculates the adjusted max battery life based on ground speed using a simplified power model.
 
@@ -73,7 +75,7 @@ def calculate_adjusted_max_battery_life(drone_type: DroneType, ground_speed: flo
       of 2:1 (hover to drag) is chosen as a reasonable starting point for many quadcopters
       at moderate speeds.
     - Wind and temperature are not factored in.
-    
+
     Simplified power model: P_total = P_hover + k * v^2
     Energy E = P_total * T
     So, T = E / (P_hover + k * v^2)
--- a/src/backend/packages/drone-flightplan/drone_flightplan/drone_type.py
+++ b/src/backend/packages/drone-flightplan/drone_flightplan/drone_type.py
@@ -30,10 +30,7 @@ DRONE_SPECS = {
         # 1/1.3-inch CMOS
         # 4:3 (or 16:9 cropped)
         # Testing speed is in kmph
-        "max_battery_life_minutes": {
-            "quoted_value": 34,
-            "testing_speed": 21.6
-        },
+        "max_battery_life_minutes": {"quoted_value": 34, "testing_speed": 21.6},
         "sensor_height_mm": 7.2,
         "sensor_width_mm": 9.6,
         "equiv_focal_length_mm": 24,
@@ -42,10 +39,7 @@ DRONE_SPECS = {
     DroneType.DJI_AIR_3: {
         # 1/1.3-inch CMOS
         # 4:3 (or 16:9 cropped)
-         "max_battery_life_minutes": {
-            "quoted_value": 46,
-            "testing_speed": 28.8
-        },
+        "max_battery_life_minutes": {"quoted_value": 46, "testing_speed": 28.8},
         "sensor_height_mm": 7.2,
         "sensor_width_mm": 9.6,
         "equiv_focal_length_mm": 24,
@@ -54,10 +48,7 @@ DRONE_SPECS = {
     DroneType.DJI_MINI_5_PRO: {
         # 1-inch CMOS
         # 4:3 (or 16:9 cropped)
-         "max_battery_life_minutes": {
-            "quoted_value": 36,
-            "testing_speed": 21.6
-        },
+        "max_battery_life_minutes": {"quoted_value": 36, "testing_speed": 21.6},
         "sensor_height_mm": 9.6,
         "sensor_width_mm": 12.8,
         "equiv_focal_length_mm": 24,
@@ -66,10 +57,7 @@ DRONE_SPECS = {
     DroneType.POTENSIC_ATOM_2: {
         # 1/2-inch CMOS
         # 4:3 (or 16:9 cropped)
-        "max_battery_life_minutes": {
-            "quoted_value": 32,
-            "testing_speed": 18
-        },
+        "max_battery_life_minutes": {"quoted_value": 32, "testing_speed": 18},
         "sensor_height_mm": 4.80,
         "sensor_width_mm": 6.40,
         "equiv_focal_length_mm": 26,


src/backend/packages/drone-flightplan/drone_flightplan/drone_type.py
```diff
@@ -18,10 +18,20 @@ class DroneType(StrEnum):
...
+# maximum speeds.
+# Results were estimated based on simulating a moderate drag force on the drone during flight
+# at 11.5 m/s. Will need testing to see how realistic these speeds are.
+# NB - mention this in some kind of disclaimer when suggesting to user that their
+# task should be split
 DRONE_SPECS = {
     DroneType.DJI_MINI_4_PRO: {
         # 1/1.3-inch CMOS
         # 4:3 (or 16:9 cropped)
+        "max_battery_life_minutes": 18,
```
spwoodcock: How is the Max battery calculated here? DJI advertise closer to 30 mins per battery, but I think based on field experience, in a good temperature range, we can probably expect more like 25 mins.

Of course battery degradation over time matters too, but we can address that later.

I think maybe we should consider closer to 25 mins if possible?

src/backend/packages/drone-flightplan/drone_flightplan/drone_type.py
```diff
@@ -38,6 +49,7 @@ class DroneType(StrEnum):
     DroneType.DJI_MINI_5_PRO: {
         # 1-inch CMOS
         # 4:3 (or 16:9 cropped)
+        "max_battery_life_minutes": 16,
```
spwoodcock: The mini 5 pro advertises a bigger battery than the mini 4 pro in theory, but I dont think there is much difference between them in reality. I would be surprised if the 5 had less battery than the 4 though

src/backend/packages/drone-flightplan/drone_flightplan/drone_type.py
stanleypliu: The maximum battery I calculated using a simulated average drag force each drone would expect to encounter flying at a speed of 11.5 m/s. I just wanted to be safe, but I can revisit the calculations and see if I can revise these upwards. 

Just to note, the battery lives they quote use a substantially lower speed for testing, e.g.
Screenshot 2025-11-15 at 14 16 05

For reference, 11.5 m/s (which is what we use as an estimate for ground speed in `calculate_parameters.py` is 41.4 kph

src/backend/packages/drone-flightplan/drone_flightplan/drone_type.py
stanleypliu: Good point, this doesn't make sense. Looking at this again

src/backend/packages/drone-flightplan/drone_flightplan/calculate_parameters.py
```diff
@@ -103,20 +103,11 @@ def calculate_parameters(
     side_spacing = side_photo_width - side_overlap_distance
     ground_speed = forward_spacing / image_interval
 
-    # While Mini 4 Pro can go 12m/s and Mini 5 Pro 15m/s, we cap the ground speed at 11.5 m/s to
+    # While Mini 4 Pro can go 12m/s, Atom 2 16m/s and Mini 5 Pro 15m/s, we cap the ground speed at 11.5 m/s to
```
spwoodcock: Where did you find the 16m/s stat?

8m/s is from me testing the drone personally, but interesting if Potensic advertise double that

src/backend/packages/drone-flightplan/drone_flightplan/calculate_parameters.py
```diff
@@ -103,20 +103,11 @@ def calculate_parameters(
...
     # avoid problems with the RC2 controller.
     # Speeds over 12 m/s cause the controller to change the speed to 2.5 m/s, which is too slow.
     # Keeping it below 12 m/s ensures the flight plan works correctly.
-    if ground_speed > 12 and (
```
spwoodcock, line 110: Not sure if this logic should change / is related to this PR?

src/backend/packages/drone-flightplan/drone_flightplan/drone_type.py
```diff
@@ -30,6 +38,7 @@ class DroneType(StrEnum):
     DroneType.DJI_AIR_3: {
         # 1/1.3-inch CMOS
         # 4:3 (or 16:9 cropped)
+        "max_battery_life_minutes": 33,
```
spwoodcock: In the calculate_parameters.py file we have some functions to reproduce calcs for the camera.

Could we put similar for the battery life calcs there? Currently there is no way to reproduce the stats for a new drone that we might add / its a bit of a black box.

Also pretty surprised the battery of the air 3 would be almost double the mini 4 pro, as I believe its a cheaper smaller model

src/backend/packages/drone-flightplan/drone_flightplan/drone_type.py
```diff
@@ -46,6 +56,7 @@ class DroneType(StrEnum):
     DroneType.POTENSIC_ATOM_2: {
         # 1/2-inch CMOS
         # 4:3 (or 16:9 cropped)
+        "max_battery_life_minutes": 12,
```
spwoodcock: Seems a bit low. From testing I think the potensic has at least 20-25 minutes in it

src/backend/packages/drone-flightplan/drone_flightplan/drone_type.py
```diff
@@ -60,6 +71,7 @@ class DroneType(StrEnum):
         "equiv_focal_length_mm": 24,
         "image_width_px": 4032,
     },
+    # TODO - need to find a way to calculate battery life for the Mavlink
```
spwoodcock, line 75: Thanks for the comment 🙏

Mavlink is actually a spec supported by many drones. So in future we could allow the user to upload their custom specs that we use here

src/frontend/src/components/DroneOperatorTask/DescriptionSection/DescriptionBox/index.tsx
```diff
@@ -234,6 +234,9 @@ const DescriptionBox = () => {
             data={description?.data}
           />
         ))}
+        <p className="naxatw-text-[0.75rem] naxatw-text-[#212121]">
+          *This flight time was calculated using an average ground speed of 11.5 m/s.
```
spwoodcock, line 240: For now we can assume the supported DJI drones will fly at about this speed. But going forward, can we assume this of new models?

src/backend/packages/drone-flightplan/drone_flightplan/waypoints.py
```diff
@@ -710,6 +688,31 @@ def create_waypoint(
...
+        total_distance += calculate_distance(point1, point2)
+
+    # Calculate estimated flight time
+    ground_speed = parameters["ground_speed"]
+
+    if ground_speed > 0:
+        estimated_flight_time_seconds = total_distance / ground_speed
+        estimated_flight_time_minutes = estimated_flight_time_seconds / 60
+    else:
+        estimated_flight_time_minutes = 0
```
spwoodcock: Probably makes more sense to put this above the `if` instead, in a similar style to the total distance logic. Good to be consistent 😄

spwoodcock (changes requested):

src/backend/packages/drone-flightplan/drone_flightplan/calculate_parameters.py
stanleypliu: It was on this page https://www.potensic.com/atom-2-specs.html. However if you tested this personally and found it to be 8 m/s we can always go with that?

src/backend/packages/drone-flightplan/drone_flightplan/calculate_parameters.py
stanleypliu: Because I saw that the Atom 2 had a speed of 16 m/s according to the spec, it meant that the condition to set the ground speed just for it was now invalid. That was the only drone that appeared to be slower than the others so it meant that the condition could be simplified by having the `ground_speed` set to 11.5. As per your above reply though, I can revert my changes.

src/frontend/src/components/DroneOperatorTask/DescriptionSection/DescriptionBox/index.tsx
stanleypliu: I feel we can adjust this as needed - didn't want to alter the already defined default when I saw it in the calculations but if a drone can't achieve this speed we would need to lower this. I can put a TODO here

src/backend/packages/drone-flightplan/drone_flightplan/calculate_parameters.py
spwoodcock: Yeah I think they are a bit over ambitious with that stat, maybe 10m/s a most!

I was getting 8m/s holding full throttle in one direction, but it would go a bit quicker with a tailwind being so light

src/frontend/src/components/DroneOperatorTask/DescriptionSection/DescriptionBox/index.tsx
spwoodcock: Perfect, thanks!

src/backend/packages/drone-flightplan/drone_flightplan/drone_type.py
stanleypliu: It's the other way round, Air 3 is actually larger and more expensive than the Mini 4 Pro according to your spreadsheet https://docs.google.com/spreadsheets/d/15QyC3Y0HT1-zZm3nhE_2hRYHh6y5AnwBnO32uFtwNTw/edit?gid=0#gid=0

src/backend/packages/drone-flightplan/drone_flightplan/drone_type.py
stanleypliu: Added the number of minutes the manufacturers quote, and the adjusted battery life works out to be around 20 minutes

PR #666, opened 2025-11-20
Fix task and individual project page console errors 

stanleypliu:

- [ ] 🍕 Feature
- [x] 🐛 Bug Fix
- [ ] 📝 Documentation
- [x] 🧑‍💻 Refactor
- [ ] ✅ Test
- [ ] 🤖 Build or CI

## Related Issue

N/A

## Describe this PR

- I noticed in my local development environment (and also some of the same errors in production) on a task page upon load, as well as when interacting with the map.
- The main issues seemed to stem from map layers being accessed or modified before they were ready.
- A few other minor issues were traced and fixed, related to the breadcrumbs and side effects in a `react-query` hook.
- I ended up fixing a few errors on the project page too, because `ToolTip` is also used there for the map buttons, otherwise I wouldn't include that fix in the same PR since it's outside the scope.

## Screenshots

N/A

N/A

## Review Guide

Whilst on the `dev` branch, have the console open when visiting a task page and there should be various console errors present in the sidebar. Interact with the map and you should see additional errors being triggered.
Now switch to this branch and refresh the page. These errors should now be absent.

### DIFF
--- a/src/frontend/src/components/IndividualProject/MapSection/index.tsx
+++ b/src/frontend/src/components/IndividualProject/MapSection/index.tsx
@@ -30,6 +30,7 @@ import ToolTip from '@Components/RadixComponents/ToolTip';
 import Legend from './Legend';
 import ProjectPromptDialog from '../ModalContent';
 import UnlockTaskPromptDialog from '../ModalContent/UnlockTaskPromptDialog';
+import Icon from '@Components/common/Icon';
 
 const MapSection = ({ projectData }: { projectData: Record<string, any> }) => {
   const { id } = useParams();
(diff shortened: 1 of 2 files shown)


spwoodcock:
Thanks so much for this - its something that has always bothered me about the frontend code here 😅

I'll review asap

src/frontend/src/components/common/MapLibreComponents/LocateUser/index.tsx
```diff
@@ -1,13 +1,9 @@
...
+import { useMap } from '../MapContext';
 
-interface ILocateTheUserProps {
-  map?: MapInstanceType;
-  isMapLoaded: Boolean;
-}
-
-const LocateUser = ({ map, isMapLoaded }: ILocateTheUserProps) => {
+const LocateUser = () => {
+  const { map, isMapLoaded } = useMap();
```
spwoodcock, line 6: This approach makes all of the map components less self-contained, but probably makes sense in this case (the components are likely just specific to this repo)

spwoodcock (approved):
Thanks again 😃

Lots of changes here! On first review is looks pretty promising - I haven't done any testing though.

Tempted to merge in and have this tested on dev, as I won't be at my dev env again until Tuesday

src/frontend/src/components/common/MapLibreComponents/LocateUser/index.tsx
stanleypliu: Correct they are specific to this repo, and I agree, but using context is a more standard practice.

PR #675, opened 2025-11-28
Add Sentry OTEL traces to prod

### LINKED ISSUE #667: Add Sentry OTEL traces to prod
Better logging to trace issues.

Same setup as FieldTM.

stanleypliu:

- [x] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation
- [ ] 🧑‍💻 Refactor
- [ ] ✅ Test
- [x] 🤖 Build or CI

## Related Issue

Fixes #667 

## Describe this PR

Adds OpenTelemetry logging via Sentry.

## Screenshots

N/A

N/A

## Review Guide

Add the new ENV variables to your `.env` making sure that you have the right values. Uncomment `INSTALL_MONITORING` in `compose.yaml` and rebuild your backend Docker container. Use a test script analogous to https://docs.sentry.io/platforms/python/tracing/instrumentation/opentelemetry/#:~:text=Finally%2C%20try%20it%20out%20with%3A to test it.

### DIFF
--- a/src/backend/pyproject.toml
+++ b/src/backend/pyproject.toml
@@ -69,13 +69,10 @@ docs = [
 ]
 monitoring = [
     "sentry-sdk[opentelemetry-otlp]>=2.46.0",
-    #"opentelemetry-sdk>=1.27.0",
-    #"opentelemetry-api>=1.27.0",
-    #"opentelemetry-exporter-otlp>=1.27.0",
-    "opentelemetry-instrumentation-fastapi>=0.48b0",
-    "opentelemetry-instrumentation-logging>=0.48b0",
+    "opentelemetry-instrumentation-fastapi>=0.59b0",
+    "opentelemetry-instrumentation-logging>=0.59b0",
     "opentelemetry-instrumentation-psycopg>=0.59b0",
-    "opentelemetry-instrumentation-requests>=0.48b0",
+    "opentelemetry-instrumentation-requests>=0.59b0",
 ]
 
 [tool.uv.sources]
--- a/src/backend/uv.lock
+++ b/src/backend/uv.lock
@@ -767,10 +767,10 @@ docs = [
     { name = "python-dotenv", specifier = "==1.0.1" },
 ]
 monitoring = [
-    { name = "opentelemetry-instrumentation-fastapi", specifier = ">=0.48b0" },
-    { name = "opentelemetry-instrumentation-logging", specifier = ">=0.48b0" },
+    { name = "opentelemetry-instrumentation-fastapi", specifier = ">=0.59b0" },
+    { name = "opentelemetry-instrumentation-logging", specifier = ">=0.59b0" },
     { name = "opentelemetry-instrumentation-psycopg", specifier = ">=0.59b0" },
-    { name = "opentelemetry-instrumentation-requests", specifier = ">=0.48b0" },
+    { name = "opentelemetry-instrumentation-requests", specifier = ">=0.59b0" },
     { name = "sentry-sdk", extras = ["opentelemetry-otlp"], specifier = ">=2.46.0" },
 ]
 test = [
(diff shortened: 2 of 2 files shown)


src/backend/app/config.py
```diff
@@ -28,6 +30,69 @@
...
+    SENTRY = "sentry"
+
+
+class OtelSettings(BaseSettings):
+    """Inherited OpenTelemetry specific settings (monitoring).
+
+    These mostly set environment variables set by the OTEL SDK.
+    """
+
+    FMTM_DOMAIN: Optional[str] = Field(exclude=True)
```
spwoodcock: This was from FieldTM but I think needs to be updated to match DroneTM env vars

src/backend/app/config.py
```diff
@@ -28,6 +30,69 @@
...
+    LOG_LEVEL: Optional[str] = Field(exclude=True)
+    ODK_CENTRAL_URL: Optional[str] = Field(exclude=True)
```
spwoodcock: This isn't relevant for dronetm

src/backend/app/monitoring.py
```diff
@@ -0,0 +1,26 @@
...
+    """Add OpenTelemetry FastAPI instrumentation.
+
+    Only used if environment variables configured.
+    """
+    from opentelemetry.instrumentation.fastapi import FastAPIInstrumentor
+    from opentelemetry.instrumentation.requests import RequestsInstrumentor
+
+    FastAPIInstrumentor.instrument_app(app)
+
+    # FIXME psycopg enable once linked issue addressed:
```
spwoodcock: The linked issue was actually fixed, so we could update the opentelemetry package versions and enable this code for psycopg =)

src/backend/app/config.py
stanleypliu: I've renamed it to `SITE_NAME`

src/backend/app/monitoring.py
stanleypliu: Oh true, we should fix this in FieldTM as well then

src/backend/pyproject.toml
```diff
@@ -68,14 +68,14 @@ docs = [
...
-    "opentelemetry-instrumentation-fastapi==0.48b0",
-    "opentelemetry-instrumentation-logging==0.48b0",
-    "opentelemetry-instrumentation-psycopg==0.48b0",
-    "opentelemetry-instrumentation-requests==0.48b0",
+    "sentry-sdk[opentelemetry-otlp]>=2.46.0",
+    #"opentelemetry-sdk>=1.27.0",
+    #"opentelemetry-api>=1.27.0",
+    #"opentelemetry-exporter-otlp>=1.27.0",
+    "opentelemetry-instrumentation-fastapi>=0.48b0",
+    "opentelemetry-instrumentation-logging>=0.48b0",
```
spwoodcock: Small thing, but I think all these libs are packaged together to be compatible. Worth updating all OTEL libs to latest to be sure, before we merge this 👍

.env.example
```diff
@@ -29,6 +29,11 @@ FRONTEND_HOST=${FRONTEND_HOST:-localhost}
 NODE_ODM_URL=${NODE_ODM_URL:-http://nodeodm:9900}
 DEBUG=${DEBUG:-True}
 
+## Monitoring (OpenTelemetry) ##
+MONITORING=${MONITORING:-sentry}
```
spwoodcock, line 33: Could this be changed to have no value by default please - we dont want the traces from dev / staging

.env.example
```diff
@@ -29,6 +29,11 @@ FRONTEND_HOST=${FRONTEND_HOST:-localhost}
...
+SENTRY_DSN=${SENTRY_DSN:-production_dsn}
```
spwoodcock, line 34: Also shouldn't have a default

stanleypliu:
> I already merged this, but forgot to mention the default should be monitoring disabled.
> 
> We dont want to send logs for dev / staging setups. It should be explicitly enabled by env var in production 😃

Yes that's true. So it won't actually send any dev or staging logs as is, because I included an `INSTALL_MONITORING` variable in the Dockerfile that would only install those dependencies for monitoring if uncommented. I forgot to mention that you'd need to change the build command to `docker compose build --build-arg INSTALL_MONITORING=true backend` for production builds - but I can address this and your above comments in a follow up.

PR #686, opened 2025-12-13
Add CI workflow and Justfile test commands

allisonsibrian:
## What type of PR is this? (check all applicable)

- [ ] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation
- [ ] 🧑‍💻 Refactor
- [X] ✅ Test
- [X] 🤖 Build or CI
- [ ] ❓ Other (please specify)

## Related Issue

Builds upon: #685

## Describe this PR

This draft PR creates automated testing for the backend. It enables running tests via just and also automates them via GitHub Actions on every push.
(First time working on setting up CI and testing configuration, so I used the field-tm repository and other hotosm workflows as a reference implementation, also why I am opening this as a draft PR first to address issues and add-ons.)

For the Justfile Commands: 
- Added a new module tasks/test/Justfile.
- Added the module in the root Justfile.
- Created just test backend which follows this cycle: Build -> Start (Wait) -> Migrate -> Test -> Cleanup

For CI automation: 
- Added .github/workflows/test.yml.
- custom workflow install's uv and just before running the backend tests on ubuntu.

CI Syntax Error: 
- I removed a line for this draft PR in .github/workflows/build_and_deploy.yml (MONITORING=${{ secrets.MONITORING }}) that was causing "Invalid workflow file" errors, as it seems that passing secrets inside a with block isn't allowed. (Only removed to be able to run tests)

## Review Guide
Can verify via CI or locally via just test backend:
- Verify test workflows are running

## Checklist before requesting a review

- 📖 Read the HOT Code of Conduct: <https://docs.hotosm.org/code-of-conduct>
- 👷‍♀️ Create small PRs. In most cases, this will be possible.
- ✅ Provide tests for your changes.
- 📝 Use descriptive commit messages.
- 📗 Update any related documentation and include any relevant screenshots.

PR #689, opened 2025-12-16
Fixing backend tests

allisonsibrian:

- [ ] 🍕 Feature
- [X] 🐛 Bug Fix
- [ ] 📝 Documentation
- [ ] 🧑‍💻 Refactor
- [X] ✅ Test
- [X] 🤖 Build or CI

## Related Issue

Fixes #686

## Describe this PR

This PR resolves the 403/404 errors encountered in the backend tests. It resolves database conflicts, fixes permission logic, and ensures all backend tests pass in CI.

Fixed background issue in app/projects/project_logic.py:
- Updated process_task_metrics to handle cases where Psycopg 3 returns points as a dictionary wrapper instead of a GeoJSON string. This prevents a TypeError in geojson.loads.

Fixed permission logic in app/auth/permissions.py:
- Restructured ordering for has_permission, where the system validates user.is_superuser before checking if the target object exists. This fixes 403 errors on "Create" endpoints where the object is None.

Conftest.py fixes: 
- Updated project_info to generate UUID-based project names and prevents "Project name already exists" issues. 
- Added is_superuser=True via SQL to ensure Admin role.
- Added an override for login_dependency, as this allows the permission system to accept the test user.

Other fixes: 
- Corrected URL paths in test_projects_routes and test_users_routes
- Updated test_reset_password_success to generate a valid token for the test user.

## Review Guide

Run backend tests and verify all 7 tests pass.

### DIFF
--- a/src/backend/app/users/permissions.py
+++ b/src/backend/app/users/permissions.py
@@ -122,10 +122,10 @@ class HasObjectPermission(BasePermission):
     ) -> bool:
         if not user:
             return False
-        
+
         if user.is_superuser:
             return True
-        
+
         if not obj:
             return False
 
--- a/src/backend/tests/conftest.py
+++ b/src/backend/tests/conftest.py
@@ -40,7 +40,7 @@ async def auth_user(db) -> AuthUser:
             name="admin",
             profile_img="",
             role=UserRole.PROJECT_CREATOR,
-            is_superuser = True
+            is_superuser=True,
         ),
     )
     db_user.is_superuser = True


spwoodcock:
Nice, fantastic to see the tests operational!

I can't seem to comment on specific lines via the github app currently, so will review all together.

1
For the is_superuser check, if user is None, then checking user.is_superuser first would AttributeError no? 

2
Also for the db.execute to update the user to a superuser, couldn't the user initially be set with the is_superuser field set to True, instead of having to update after?

allisonsibrian:
> For the is_superuser check, if user is None, then checking user.is_superuser first would AttributeError no?
> 
> 2 Also for the db.execute to update the user to a superuser, couldn't the user initially be set with the is_superuser field set to True, instead of having to update after?

1. Thank you for catching this. I didn't realize when I swapped the blocks, but I'll update this!
2. You're right that initially setting it is better. I originally added the SQL to handle already existing test users that didn't have any Admin privileges, but it's much simpler to set AuthUser to accept is_superuser=True and still keep db_user.is_superuser = True to handle those cases.

Issue #671, opened 2025-11-26
Add E2E/integration tests for critical user flows


stanleypliu:
Thoughts on adding E2E tests (Cypress etc) to test critical user flows like logging in/signup, creating a project, locking and accessing a task and so on? I notice there aren't any at the moment. I don't think it would impact deployment times that much either.

spwoodcock:
We have a contract ongoing with a huge refactor to the photo upload process currently.

But mostly until Feb we are basically just struggling along with my limited time and any volunteer contribution where possible 🙏

In Feb a new lead dev starts and development will increase rapidly - lots of demand and loads to do 😁

PR #693, opened 2025-12-23
Smoke and endpoint tests

stanleypliu:

- [ ] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation
- [x] 🧑‍💻 Refactor
- [x] ✅ Test
- [x] 🤖 Build or CI

## Related Issue

Part 1 of #671 

## Describe this PR

Add smoke tests for the various endpoints in this repository. Not every endpoint is tested, only those which were deemed critical and shouldn't vary in the average feature cycle, e.g. user routes. 
Test suite took a couple minutes to run on my local environment, including setup and teardown of containers so it doesn't seem overly impactful on performance.  

## Screenshots

N/A

N/A

## Review Guide

Run `just test backend` with the latest Docker backend container running.

### DIFF
--- a/compose.test.yaml
+++ b/compose.test.yaml
@@ -12,26 +12,13 @@ networks:
 
 services:
   backend:
-    image: ghcr.io/hotosm/drone-tm/backend:debug
+    extends:
+      file: compose.yaml
+      service: backend
     build:
       context: src/backend
-      dockerfile: Dockerfile
       args:
         INSTALL_TEST: "true"
-    ports:
-      - "8001:8000" # Avoid conflict with main backend on 8000
-    volumes:
-      - ${PROJECT_DIR:-.}/src/backend/pyproject.toml:/project/src/backend/pyproject.toml:ro
-      - ${PROJECT_DIR:-.}/src/backend/app:/project/src/backend/app:ro
-      - ${PROJECT_DIR:-.}/src/backend/tests:/project/src/backend/tests:ro
-      - ${PROJECT_DIR:-.}/src/backend/packages/drone-flightplan/drone_flightplan:/opt/python/lib/python3.11/site-packages/drone_flightplan:ro
-      - frontend-html:/project/src/backend/frontend_html
-    env_file: .env
-    extra_hosts:
-      - "localhost:host-gateway"
-    networks:
-      - dtm-network
-    restart: unless-stopped
     environment:
       MONITORING: ""
       LOG_LEVEL: "INFO"
@@ -41,7 +28,7 @@ services:
       db:
         condition: service_healthy
       minio:
-        condition: service_healthy
+        condition: service_started
       redis:
         condition: service_started
       nodeodm:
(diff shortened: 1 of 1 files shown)


compose.test.yaml
```diff
@@ -13,13 +12,26 @@ networks:
 
 services:
   backend:
-    extends:
```
spwoodcock, line 16: How come the `extends` based syntax doesn't work here?

It prevents duplication & forgetting to update compose.test.yaml when we update compose.yaml - avoiding a mismatch in future.

Is it possible to make this PR while keeping the `extends` syntax? 🙏

src/backend/Dockerfile
```diff
@@ -57,19 +57,18 @@ ARG INSTALL_TEST=false
...
-    fi
-
-    if [ "$INSTALL_TEST" = "true" ]; then
-        EXTRA_GROUPS="$EXTRA_GROUPS --group test"
-    fi
-
-    uv sync --locked --no-editable --no-dev $EXTRA_GROUPS
-EOT
+    --mount=type=bind,source=uv.lock,target=uv.lock \
+    sh -c ' \
```
spwoodcock: I think this new syntax is less readable. What value did it bring?

src/backend/app/drones/drone_routes.py
```diff
@@ -20,79 +15,6 @@
 )
 
 
-@router.get("/", response_model=list[drone_schemas.DroneOut])
```
spwoodcock, line 23: Cleaning these up would also mean cleaning up and removing the `Drones` database model / table.

I'm not sure we should take that decision from within this PR, but instead make that another issue.

Going to https://dronetm.org/api/drones/ currently only returns:

```json
[{"id":1,"model":"DJI Mini 4 Pro"}]
```

I think the idea was to put drone models in the db, instead of hardcoding into the drone-flightplan module, but that proves difficult if we want to decouple them.

We may eventually repurpose the `drones` table to include user provided information for drones - particularly custom drones using mavlink etc.

Let's keep these endpoints for now, and consider this in a new issue 😄

src/backend/tests/test_drones_routes.py
```diff
@@ -4,26 +4,15 @@
...
+async def test_get_drone_altitude_by_country(client):
+    """Test getting drone altitude by country."""
+    country = "canada"
+    response = await client.get(f"/api/drones/drone-altitude/{country}/")
     assert response.status_code == HTTPStatus.OK
 
     return response.json()
 
 
-@pytest.mark.asyncio
```
spwoodcock, line 15: I agree this test is a bit pointless 👍

spwoodcock (changes requested):
Looks great! Thanks 😃 

I made a few comments to address, but otherwise it's pretty much there

compose.test.yaml
stanleypliu: I couldn't get `just test backend` to work properly locally as the ports were already in use. I'll try another approach though.

src/backend/Dockerfile
stanleypliu: ~~I couldn't install `pytest` properly which was the reasoning behind my change.~~ Sorry, confused the issue with another, I made this change because `set` wasn't recognised as a valid keyword by Docker inside the heredoc.

To try and fix this whilst preserving the old structure, I added a line at the top of this file which enables BuildKit  - there do seem to be downsides using this though, and I don't know enough about Docker (yet) to be able to evaluate them properly. Other than that though, there doesn't seem to be any way to keep `set` and not have the build process error when running it on the command line - I already have `buildkit` enabled for Docker Desktop so don't have this issue when starting builds there.

src/backend/app/drones/drone_routes.py
stanleypliu: Noted! I did recall that you wanted to get rid of these though (or at least not store them in the DB), so I thought I'd go ahead and make a start on removing the unused endpoints.

compose.test.yaml
stanleypliu: To elaborate on my reply above: my reasoning was to allow the tests to run without having to stop running containers for the application but it was too excessive a change just to run them locally!

spwoodcock:
Excellent work - this is a great improvement to help ensure these endpoints dont break further down the line 🙌

PR #698, opened 2026-01-16
Removing flight tails from uploaded imagery

### LINKED ISSUE #599: Remove flightplan tails during post processing
## Problem Summary

- We encountered some issues during processing of Freetown imagery.
- The first 20-30 images are taken from the take off point to the first waypoint position.
- This is because we need to use the interval photo setting from the take off point (detailed waypoint missions with photo capture at each point would not have this issue, but in order to fit the whole mapping area in the DJI Fly app, we simplify the waypoints to 'waylines' - essentially waypoints for the start and end of each line to fly, plus a photo interval).
- These tails should be removed in an automated way, prior to final imagery processing (or perhaps even prior to imagery upload if at all possible, but less likely).

## Documenting points made by @smathermather:

- "Tilt in the models themselves... The following is a difference map with a reference low resolution terrain model, and we can see some different variance from the low resolution model in the corners of 058 and 056 and probably others":

- "When we've seen this previously, we've selectively removed the images that are collected en route and the tilt goes away, but this is a fairly large subset of the total data (~30k images), so probably good to take a first stab at automated removal."

- "If we look at our flights, we can see these little tails from the takeoff point to the main part of the flight plan":

- "So, we need to identify and remove these and hope this fixes our issue"

## Possible solution from @smathermather 

- "So, we pick each 2 photos and calculate a line between them so we can get the azimuth of that line":

- "But, we probably only need to filter on the first 30 images, which will make any angle/azimuth based filtering safer, so we subset and look at that angle":

And remove anything that isn't near 90-degrees:

It's not perfect, but probably good enough:

- SQL algorithm that could achieve this:

```sql
DROP TABLE IF EXISTS freetown.freetown_flight_lines_start;
CREATE TABLE freetown.freetown_flight_lines_start AS
	WITH beginning AS (
		SELECT
		  * 
		FROM (
		  SELECT
			ROW_NUMBER() OVER (PARTITION BY flight ORDER BY datetime) AS row,
			t.*
		  FROM
			freetown.freetown_flights t) x
		WHERE
		  x.row <= 30
		),
	ordered AS (
		SELECT
			id,
			flight,
			datetime,
			geom,
			LAG(geom) OVER (PARTITION BY flight ORDER BY datetime) AS previous_geom,
			LAG(datetime) OVER (PARTITION BY flight ORDER BY datetime) AS previous_datetime
		FROM
			beginning
		),
	lines AS (
		SELECT id, flight, previous_datetime, datetime, degrees(ST_Azimuth(previous_geom, geom)) AS azimuth, 
			ST_MakeLine(previous_geom, geom) AS geom FROM ordered
		)
	SELECT * FROM lines WHERE geom IS NOT NULL;
```

allisonsibrian:

- [X] 🍕 Feature
- [X] 🐛 Bug Fix
- [ ] 📝 Documentation
- [ ] 🧑‍💻 Refactor
- [ ] ✅ Test
- [ ] 🤖 Build or CI

## Related Issue

Fixes #599 

## Describe this PR

This draft PR implements a system to detect and flag flightplan tails during take-off and landing phases of the flight trajectory. This ensures that any transit imagery is excluded from later ODM processing. 

The implementation encompasses:
- A search limit: The search boundaries are scaled based on the batch size. 
          a. The search is lower-bounded at 30 images and upper-bounded at 60 images (or roughly 1/3 of the batch size).
          b. For small batches that are less than 30 images, I added an extra search limit that ensures the limit never exceeds the amount of available data.
- A moving baseline:  Instead of a fixed 90-degree filter, this PR uses a moving average baseline of the takeoff/landing trajectory to account for minor flight changes
- A 45-degree deviation trigger: A 45-degree deviation is used as a suspected turn
- Confirmation in change in direction: To prevent any false positives from flight changes, there is a 5-image look-ahead(takeoff) and look-behind window (landing) to confirm if the trajectory remains stable within a 10-degree tolerance and actually detects a turn.

## Review Guide
- src/backend/app/images/image_logic.py: Contains the trajectory analysis, running baseline, and confirmation turn logic.
- src/backend/app/tasks/task_logic.py: Added the logic into the process_batch_images workflow. Tails are flagged as REJECTED immediately before images are moved to task folders.

I created a quick logic test and executed it within the Docker container environment. I was able to confirm: 
- Azimuthal difference calculations.
- Confirmation of turns.
- Correct flagging of takeoff indices in a 15-image test trajectory.

I would like to test this out with actual imagery.  What would be the best way to test this change using test datasets within the batch workflow? 

### DIFF
--- a/src/backend/app/images/image_logic.py
+++ b/src/backend/app/images/image_logic.py
@@ -380,27 +380,24 @@ async def get_images_by_project(
     return [ProjectImageOut(**row) for row in results]
 
 
-def _calculate_angular_difference(
-        degree1: float, degree2: float
-        ) -> float: 
+def _calculate_angular_difference(degree1: float, degree2: float) -> float:
     """
-   Calculates the shortest angular difference between two angles. 
+    Calculates the shortest angular difference between two angles.
 
-   Ensures the difference accounts for 360-degree wrap-around.
+    Ensures the difference accounts for 360-degree wrap-around.
 
-   Returns: 
-        float: Absolute difference in degrees (0 to 180)
+    Returns:
+         float: Absolute difference in degrees (0 to 180)
     """
     angular_difference = abs(degree1 - degree2)
     if angular_difference > 180:
-        angular_difference = 360 - angular_difference 
+        angular_difference = 360 - angular_difference
     return angular_difference
 
-def _calculate_circular_mean(
-        degree1: float, degree2: float
-    ) -> float: 
+
+def _calculate_circular_mean(degree1: float, degree2: float) -> float:
     """
-    Calcultes the mean between two angles, accounting for 360-degree wrap around. 
+    Calculates the mean between two angles, accounting for 360-degree wrap around.
     """
     rad1 = math.radians(degree1)
     rad2 = math.radians(degree2)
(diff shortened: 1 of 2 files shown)


src/backend/app/images/image_logic.py
```diff
@@ -354,3 +355,216 @@ async def get_images_by_project(
...
+            takeoff_tails_indices.append(i)
+            continue
+
+        azimuth_difference = _calculate_angular_difference(
+            takeoff_baseline, current_azimuth
+        )
+        # A 45-degree change in baseline indicates a change in direction during the flight.
+        if azimuth_difference < 45:
+            takeoff_tails_indices.append(i)
+            takeoff_baseline = (takeoff_baseline + current_azimuth) / 2
```
spwoodcock: I think for this we must use a circular mean:

```python
def circular_mean(deg1: float, deg2: float) -> float:
    r1 = math.radians(deg1)
    r2 = math.radians(deg2)
    x = math.cos(r1) + math.cos(r2)
    y = math.sin(r1) + math.sin(r2)
    return (math.degrees(math.atan2(y, x)) + 360) % 360
```

For example if we have baseline 350, but current 10, the mean should be 0, but running through the current logic would give 180. Please correct me if I'm wrong 🙏

src/backend/app/images/image_logic.py
```diff
@@ -354,3 +355,216 @@ async def get_images_by_project(
...
+    Returns:
+         float: Absolute difference in degrees (0 to 180)
+    """
+    angular_difference = abs(baseline - current_azimuth)
+    if angular_difference > 180:
+        angular_difference = 360 - angular_difference
+    return angular_difference
+
+
+def _confirm_turn_made(project_list: list, image_index: int, steps: int) -> bool:
```
spwoodcock: Perhaps this should be `_confirm_stable_heading`, as the logic actually confirms stability of the flight, and returns False if there is a turn

src/backend/app/images/image_logic.py
```diff
@@ -354,3 +355,216 @@ async def get_images_by_project(
...
+    takeoff_tails_indices = []
+
+    landing_loop = range(project_length - 1, project_length - search_limit, -1)
+    landing_baseline = None
+    landing_tails_indices = []
+
+    # 1. Check for flight tails that occur during takeoff.
+    for i in takeoff_loop:
+        # Stationary images are flagged as tails as they provide unreliable azimuths.
+        if project_image_results[i]["distance_moved"] < 1.0:
```
spwoodcock: This is a really great idea, but drones often pause briefly at the end of a line to re-orient. The photo they take at the end of the line might be discounted here? Perhaps we need some logic to account for if the image is at the start or end of the line, it could be treated differently?

The same for the landing check

src/backend/app/images/image_logic.py
```diff
@@ -354,3 +355,216 @@ async def get_images_by_project(
...
+        if takeoff_baseline is None:
+            takeoff_baseline = current_azimuth
+            takeoff_tails_indices.append(i)
+            continue
+
+        azimuth_difference = _calculate_angular_difference(
+            takeoff_baseline, current_azimuth
+        )
+        # A 45-degree change in baseline indicates a change in direction during the flight.
+        if azimuth_difference < 45:
```
spwoodcock, line 574: Let's keep the 45 assumption for now, but definitely add a note that this may need to be tuned.

There could be a few deviations from this, for example during terrain following missions, or if it's really windy - the drone won't follow the expected flight paths in a set grid

spwoodcock (changes requested):
accidentally approved - a couple of small changes to make 👍 😄

src/backend/app/images/image_logic.py
allisonsibrian: Ah you're right! thanks for catching this and will implement this to handle the baseline averaging!

src/backend/app/images/image_logic.py
allisonsibrian: Thanks for pointing this out! 
I can refactor to defer flagging and look for a specific index where the mission begins. If a confirmed turn is found, images that came before that specific index are flagged as tails. If no turn occurred, then the images are preserved. This ensures that the brief pauses the drone takes are not misidentified as tails.

allisonsibrian:
> I'm making a few test datasets to make this easier. In particular, when done, https://github.com/spwoodcock/drone-testdata-agung-2 will have flight tails that can be removed / tested with.

Nice, thanks, Sam!

PR #747, opened 2026-03-26
Add dev server deployment

hg1g:
Add workflow and compose for dev.drone.hotosm.org

### DIFF
--- a/.github/workflows/deploy-dev.yml
+++ b/.github/workflows/deploy-dev.yml
@@ -38,15 +38,13 @@ jobs:
       - name: Build and push frontend
         run: |
           docker build -t ghcr.io/${{ env.IMAGE_PREFIX }}/frontend:dev \
+            -f src/frontend/Dockerfile \
             --target prod \
             --build-arg VITE_AUTH_PROVIDER=${{ vars.AUTH_PROVIDER || 'legacy' }} \
             --build-arg VITE_HANKO_URL=${{ vars.HANKO_API_URL || 'https://dev.login.hotosm.org' }} \
             --build-arg VITE_FRONTEND_URL=https://dev.drone.hotosm.org \
             --build-arg VITE_API_URL=https://dev.drone.hotosm.org/api \
-            --build-arg API_URL=https://dev.drone.hotosm.org/api \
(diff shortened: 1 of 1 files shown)


src/frontend/Dockerfile
```diff
@@ -19,6 +19,10 @@ ENTRYPOINT ["/bin/sh", "-c", "pnpm run start"]
 
 # Generate frontend build files
 FROM base AS build
+ARG VITE_AUTH_PROVIDER
```
spwoodcock, line 22: All the drone-tm images actually use runtime config, not build config. Does it work using the stock images build instead of modifying?

.github/workflows/deploy-dev.yml
spwoodcock, line 1: Seems like overkill considering we will delete dev after this work is done, but happy to go ahead like this!

(just need to remember to delete the workflow after 🤞)

compose.dev.yaml
spwoodcock, line 1: This could be deployed using compose.sub.yaml, which covers dev, stage, main (old prod)

src/frontend/Dockerfile
hg1g: You're right! I checked and drone-tm uses runtime config via docker-entrypoint.sh. The VITE_* ARGs aren't needed for the current dev branch. We can remove them and use stock images. When we merge the Hanko functionality, we'll add the new vars to the runtime config instead.

compose.dev.yaml
hg1g: got it, I can use compose.sub.yaml with the right env vars. I'll set up the .env with DOMAIN=dev.drone.hotosm.org, GIT_BRANCH=dev, etc. Do you have an S3 bucket we can use, or should I set up MinIO separately?

hg1g: @spwoodcock I think I'll go with manual deploy. I have server access and can use compose.sub.yaml with the right env vars. No need to merge/revert or update the PR. What do you think?

spwoodcock: Sounds good to me - the least hassle option! 🙏

Let me know if I can help =)

There are justfile commands to handle deploy, using envsubst and compose

PR #752, opened 2026-03-29
fix(#735): integrate geojson-aoi-parser for robust AOI parsing

### LINKED ISSUE #735: Improve parsing of project area geojson using geojson-aoi-parser
- Parsing FeatureCollection - the default format exported by QGIS - fails currently for DroneTM.
- The package [geojson-aoi-parser](https://github.com/hotosm/geojson-aoi-parser) was made to standardise the AOI parsing across multiple projects, so it should be integrated.
- Users can then upload geojsons in any format & we can extract out a Polygon from it.

**Edit**
I got a featcol like this from a user:
[KYC Mathare Villages.json](https://github.com/user-attachments/files/25792772/KYC.Mathare.Villages.json)
Image

Obvious this causes an error in DroneTM as there are multiple geoms.
geojson-aoi-parser already has inbuilt 'merge' logic that would convex hull these geoms, not resulting in an error.

m4cd4r4:
## Summary

Fixes #735

Multi-feature FeatureCollections (the default export from QGIS) previously failed or silently used only the first feature when uploaded as a project AOI. This integrates [geojson-aoi-parser](https://github.com/hotosm/geojson-aoi-parser) for standard PostGIS-based normalisation across the two affected endpoints.

### What changed

- **New dependency:** `geojson-aoi-parser>=0.3.5` added to `pyproject.toml`
- **New `normalize_aoi` dependency** in `project_deps.py` - reads uploaded GeoJSON, normalises via `parse_aoi_async`, merges multiple features into a single polygon using `unary_union` + convex hull fallback
- **`preview-split-by-square`** and **`/waypoints/`** endpoints now accept any valid GeoJSON type (Polygon, MultiPolygon, Feature, FeatureCollection, GeometryCollection) and merge multi-feature inputs
- **`TaskSplitter.geojson_to_shapely_polygon`** merges multiple features instead of raising `ValueError("cannot contain multiple geometries")`
- **`process_waypoints_and_waylines`** accepts pre-normalised dict instead of raw UploadFile (validation moved to dependency layer)

### Specific user scenario fixed

The [KYC Mathare Villages.json](https://github.com/hotosm/drone-tm/issues/735#issue-2882891266) from the issue (a FeatureCollection with multiple polygons) would now be parsed, merged, and split into tasks correctly.

## Test plan

- [ ] Unit tests for `TaskSplitter.geojson_to_shapely_polygon` (single, multi-adjacent, disjoint, empty inputs)
- [ ] Integration test for `preview-split-by-square` with a multi-feature FeatureCollection
- [ ] Verify existing single-polygon uploads still work unchanged
- [ ] CI passes (existing test suite + new tests)

### DIFF
--- a/src/backend/app/projects/project_deps.py
+++ b/src/backend/app/projects/project_deps.py
@@ -11,8 +11,6 @@ from geojson_aoi import parse_aoi_async
 from loguru import logger as log
 from psycopg import Connection
 from psycopg.rows import dict_row
-from shapely.geometry import mapping, shape
-from shapely.ops import unary_union
 
 from app.db import database
 from app.models.enums import HTTPStatus
@@ -98,8 +96,8 @@ async def normalize_aoi(
     containing one merged Polygon.
 
     Uses geojson-aoi-parser for PostGIS-based normalisation (z-removal,
-    polygon orientation, type coercion), then merges multiple features
-    into a single polygon via shapely if needed.
+    polygon orientation, type coercion) with merge=True to combine
+    multiple features into a single polygon via ST_UnaryUnion.
     """
     file_ext = (project_geojson.filename or "").rsplit(".", 1)[-1].lower()
     if file_ext not in ("geojson", "json"):
@@ -108,7 +106,7 @@ async def normalize_aoi(
     content = await project_geojson.read()
 
     try:
-        featcol = await parse_aoi_async(db, content)
+        featcol = await parse_aoi_async(db, content, merge=True)
     except Exception as e:
         log.warning(f"geojson-aoi-parser failed: {e}")
         raise HTTPException(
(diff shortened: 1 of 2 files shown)


src/backend/app/projects/project_deps.py
```diff
@@ -81,3 +85,49 @@ async def geojson_upload(
...
+    db: Annotated[Connection, Depends(database.get_db)],
+    project_geojson: UploadFile = File(...),
+) -> geojson_lib.FeatureCollection:
+    """Normalise an uploaded AOI GeoJSON to a single-polygon FeatureCollection.
+
+    Accepts any GeoJSON type (Polygon, MultiPolygon, Feature,
+    FeatureCollection, GeometryCollection) and returns a FeatureCollection
+    containing one merged Polygon.
+
+    Uses geojson-aoi-parser for PostGIS-based normalisation (z-removal,
```
spwoodcock, line 98: I believe geojson-aoi-parser has a merge param that would avoid all of the shapely logic entirely 😄

src/backend/app/projects/project_routes.py
```diff
@@ -280,42 +280,30 @@ async def upload_project_task_boundaries(
...
+    project_shape = shape(aoi["features"][0]["geometry"])
+
     if no_fly_zones:
         no_fly_content = await no_fly_zones.read()
         no_fly_zones_geojson = geojson.loads(no_fly_content)
         no_fly_shapes = [
             shape(feature["geometry"]) for feature in no_fly_zones_geojson["features"]
         ]
         no_fly_union = unary_union(no_fly_shapes)
+        project_shape = project_shape.difference(no_fly_union)
```
spwoodcock: In a similar vein to geojson-aoi-parser, I wonder if we could use https://postgis.net/docs/ST_Difference.html via a db call here?

PR #780, opened 2026-04-10
Separate out image classify step, resilient to restarts, per project not per batch

### LINKED ISSUE #772: What happens to images when they are 'unmatched' but not rejected?
Good example of a project:

https://drone.hotosm.org/projects/settlement-profiling-drone-mapping-phase-1-a-collaboration-by-sdi-kenya-hot-esa-hub-and-kenya-flying-labs-260320260657

Image

Here we have multiple 'unmatched' images, but they aren't explicitly rejected images, so are still present in the DB records, and presumably in the S3 staging area too. Two questions:

1. Ideally the user would have a way to manually match the 'unmatched' photos. I.e. it's just outside of task 1, so 'Assign to xxx' where the user selects a task area and we place it within the imagery for that task.
2. If the imagery is truly unmatched and not belonging in any task area, i.e. it's a wasted photo. We should have a way to delete / reject it.

Currently, what we have is:
- The 'unmatched' image records are in the db, so they displayed on the map in yellow.
- Presumably the imagery still exists in the S3 staging area `user-uploads`.
- But for now, there is no way to do anything with this unmatched imagery - it just shows on the map and is a bit useless.
- During the classification, we could view these images in a separate section for photos with 'Issues'. Once the classification is done, and the user goes to the Verify Imagery dialog, we can no longer see the section of photos with 'Issues'.

spwoodcock:

- [ ] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation
- [x] 🧑‍💻 Refactor
- [ ] ✅ Test
- [ ] 🤖 Build or CI

## Related Issue

Fixes https://github.com/hotosm/drone-tm/issues/772

## Describe this PR

- A few people reported not being able to classify imagery after successful upload.
- I guest upload can take a while and anything could happen in-between.
- The classify step was moved to it's own dialog, no more steppers.
- This way each of the steps can be resumed at any point.
- Also, removed the idea of **batch** classifying imagery. We simply upload imagery whenever - dups are caught. This imagery is marked as STAGED, for when going to the classify step: all STAGED images are loaded up for classifying. (the batch ID is still recorded for provenance, but doesn't impact image upload / classifying).
- A fair bit of this was assisted by Opus 4.6 LLM. The refactor was manually reviewed and tweaked multiple times. Added some good tests that all pass. Also had Codex 5.4 do a few sanity check passes, that managed to catch a couple of edge cases nicely too.
- I have tested all of this locally and it works well - spent a while testing many scenarios - I think it's better than whats on prod, so ready to push and test out!
- This reduced down code in the workflows by about 1000 lines... significant simplification without dealing with 'batches'. A good chunk of new code is for the additional tests too.


spwoodcock:
Note that https://github.com/hotosm/drone-tm/pull/781 was merged into this, so the PR includes the changes for ingesting existing imagery (manually copied to S3).

PR #792, opened 2026-04-18
feat(images): replace global blur detection with grid-based sharpness and HSV terrain classification

### LINKED ISSUE #789: Imagery token over bodies of water are identified as 'blurry'
- Not sure what to do with this imagery really - good example from Freetown below:

Image

- First off, this creek in the middle of the city should probably have been removed from the AOI? How useful is imagery there? Maybe we can identify water on the map and mark the same way as no-fly zones are.
- Second, the imagery is marked fine in lots of the northern part of the creek:

Image

the imagery is interesting, but looks generally ok:

Image

then we have a lot of images marked 'blurry' in the south part:

Image

The imagery actually looks similar to above, but says the shapeness is low?

Image

Image

I don't think this is a fair assessment, but a strange artifact of trying to do the detection of images of water / a creek.

We should decide how much we want to focus on finding out why though - as mentioned above, imagery over water isn't particularly useful for our primary use cases.

FeezyHendrix:
## Summary

- Replace single-pass Laplacian variance blur detection with 4x4 grid-based sharpness scoring (75th percentile of textured cells), preventing false "blurry" rejection on low-texture terrain like water, sand, and snow
- Add HSV color-based terrain type classification (water, sand, snow_ice, bare_soil, dense_vegetation, urban, vegetation, mixed) robust to lighting/brightness conditions
- Detect dirty/muddy water via texture uniformity regardless of color hue
- Store `terrain_type` in database (`DbProjectImage`) with Alembic migration, surface it in backend API responses and frontend image review UI (thumbnail badges, map popups, image modal)

### Backend
- `app/images/image_classification.py` — Grid sharpness scoring, HSV terrain classifier, dirty water detection, DB persistence updates
- `app/db/db_models.py` — Added `terrain_type` column to `DbProjectImage`
- `app/images/image_schemas.py` — Added `terrain_type` to `ProjectImageOut`
- `app/migrations/versions/add_terrain_type.py` — Alembic migration

### Frontend
- `src/frontend/src/services/classification.ts` — `terrain_type` on TypeScript interfaces
- `src/frontend/src/components/.../ImageReview.tsx` — Terrain badges on thumbnails, popups, and modal

### Tests
- `tests/test_image_quality_classification.py` — 15 tests covering grid sharpness, all terrain types (including bright water and dirty water), exposure, and distance

Closes #789

Signed-off-by: feezyhendrix <switchdevelop1@gmail.com>


FeezyHendrix:

spwoodcock (commented):
Love this - thank you @FeezyHendrix!

Just a few questions:
1. Did you manage to test these changes on some existing imagery? Perhaps the project that is linked in the issue, showing a river delta. I'm wondering if it manages to identify that & if we get any mis-identification, allowing images that should fail to actually pass.
2. I guess the `terrain_type` field is interesting to store, but I personally don't think it adds much value to the user. We can use it internally for the classification and either passing or failing the image, but telling the user the image is grassland, when they can clearly see that on the image, doesn't add too much. Identification of imagery by humans is probably going to be significantly better than this machine based approach. (I only mention this as it's possibly unecessary db data to store, plus means we need to do a migration). What do you think? Can you foresee any uses for this field?

Thanks again - excited to test this out! 🙏

FeezyHendrix:
Hey @spwoodcock, yes it works for detecting a body of water, and ensure to not mis-identify them.

For the second one, I think, it actually doesn't provide a user value, I was thinking of a way to possibly track the mis-identification, if the need arises, to see what the code is classifying the terrain as. If there isn't a user for it, I can remove it. 

Thanks!

spwoodcock:
Nice thanks for confirming. Including images of water is important for cities with lakes, coastlines, deltas, etc.

If the user **doesn't** want a lake or the delta included, I think it should be down to them to omit from the AOI, or via no-fly zones.

Personally I can't see a use for the terrain type though, as long as we are accurately flagging blurry imagery, but not flagging bodies of water - in which case, let's remove it 😄

(if its sand, vegetation, whatever - it won't affect the processing - we want all images of all terrain types, except if there are issues with the imagery quality)

spwoodcock (approved):

PR #803, opened 2026-05-04
feat/scripts: add a script to make a CSV catalog of EXIF/XMP fields from drone photos in S3 or in a folder

zestyping:

- [x] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation
- [ ] 🧑‍💻 Refactor
- [ ] ✅ Test
- [ ] 🤖 Build or CI

## Describe this PR

This adds a handy tool to extract EXIF/XMP fields (such as GPS latitude and longitude, camera and flight angles) from drone photos in an S3 bucket or a local folder, producing a CSV file with one row for each JPEG file.

## AI Tool Usage

- [ ] No AI tools were used

**If AI-assisted:**

- Tools used: Claude
- What was generated: Claude drafted an initial version of this script as a reference, particularly to illustrate the use of the ThreadPoolExecutor.
- What you reviewed and changed: This script was completely written by hand; no code by Claude remains.

## Review Guide

To try this script, you can run it like this:

```
    python3 catalog_drone_photos.py PATH                                         
    python3 catalog_drone_photos.py s3://BUCKET_NAME                             
    python3 catalog_drone_photos.py s3://BUCKET_NAME/PREFIX                      
``` 

## Checklist

- [x] PR is focused and small
- [ ] Tests are included or updated

### DIFF
--- a/scripts/catalog_drone_photos.py
+++ b/scripts/catalog_drone_photos.py
@@ -1,0 +1,254 @@
+#!/usr/bin/env python3
+
+"""
+Generates a CSV catalog of the drone photos in a directory or an S3 bucket,
+including file metadata and useful GIS data from EXIF and XMP tags:
+  - Timestamp (when captured)
+  - Filename
+  - Path (file path or S3 object key)
+  - File size (bytes)
+  - GPS latitude
+  - GPS longitude
+  - Absolute altitude
+  - Relative altitude
+  - Camera gimbal pitch (degrees above horizon)
+  - Camera gimbal yaw (degrees CCW from north)
+  - Flight yaw (degrees CCW from north)
+  - Flight horizontal speed (m/s)
+  - Flight vertical speed (m/s)
+
+The EXIF and XMP tags are obtained by looking at just the first 64 KiB of each
+image file, to avoid downloading the entire file from S3.
+
+Only .jpg and .jpeg files are scanned.
+
+Usage:
+    python3 catalog_drone_photos.py PATH
+    python3 catalog_drone_photos.py s3://BUCKET_NAME
+    python3 catalog_drone_photos.py s3://BUCKET_NAME/PREFIX
+
+Requirements:
+    pip3 install boto3 exifread
+"""
+
+import argparse
+import boto3
+import csv
+import exifread
+from concurrent.futures import ThreadPoolExecutor, as_completed
+from datetime import datetime, timezone
+from io import BytesIO
+import os
+import re
+import sys
+
+HEAD_BYTES = 65536
+FILE_EXTENSIONS = ['.jpg', '.jpeg']
+
+
+def list_jpegs_in_dir(path):
+    """Yields (path, size) for each JPEG file under the given path."""
+    for dirpath, dirnames, filenames in os.walk(path):
+        for filename in filenames:
+            base, ext = os.path.splitext(filename)
+            if ext.lower() in FILE_EXTENSIONS:
+                filepath = os.path.join(dirpath, filename)
+                yield filepath, os.path.getsize(filepath)
+
+
+def list_jpegs_in_s3(s3_url):
+    """Yields (s3_url, size) for each JPEG under the given S3 URL."""
+    assert s3_url.startswith('s3://')
+    bucket, prefix = (s3_url[5:].split('/', 1) + [''])[:2]
+    s3 = boto3.client('s3')
+    paginator = s3.get_paginator('list_objects_v2')
+    for page in paginator.paginate(Bucket=bucket, Prefix=prefix):
+        for obj in page.get('Contents', []):
+            base, ext = os.path.splitext(obj['Key'])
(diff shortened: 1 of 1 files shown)


zestyping:

zestyping:
recheck

scripts/catalog_drone_photos.py
```diff
@@ -0,0 +1,254 @@
...
+    """Yields (s3_url, size) for each JPEG under the given S3 URL."""
+    assert s3_url.startswith('s3://')
+    bucket, prefix = (s3_url[5:].split('/', 1) + [''])[:2]
+    s3 = boto3.client('s3')
+    paginator = s3.get_paginator('list_objects_v2')
+    for page in paginator.paginate(Bucket=bucket, Prefix=prefix):
+        for obj in page.get('Contents', []):
+            base, ext = os.path.splitext(obj['Key'])
+            if ext.lower() in FILE_EXTENSIONS:
+                yield f's3://{bucket}/{obj['Key']}', obj['Size']
```
spwoodcock: The quoting breaks the f-string here:
wrap in double quote --> `yield f"s3://{bucket}/{obj['Key']}", obj['Size']`

spwoodcock (changes requested):
Love this!

Tiny syntax thing, otherwise lgtm 😄

scripts/catalog_drone_photos.py
zestyping: Ah, the nested quotes work in Python 3.12 but not lower versions.  I originally had this written differently, but it was force-upgraded by pyupgrade.  I'll try to find a syntax that both pyupgrade and Python < 3.12 will accept.

@spwoodcock What minimum version of Python do we expect users to have?

zestyping: @spwoodcock And what Python version are we forcing code to with pyupgrade?  Maybe we shouldn't use pyupgrade to force upgrade Python beyond versions that we expect everyone to have?  Or maybe we don't need pyupgrade at all — do we actually get a lot of benefit from it?

spwoodcock: Ah that's a pain!

We are currently pinned to 3.11 minimum: https://github.com/hotosm/drone-tm/blob/9c99ac98da4f16b1cf97e638f6577d344a712756/src/backend/pyproject.toml#L47

I think the pyupgrade tool should be respecting that 🤞 

If we bump the required Python version to 3.12, then pyupgrade will run to upgrade to 3.12 syntax - at least it should!

spwoodcock: Was it re-written by pyupgrade, or possibly by `ruff` instead? Hard to tell. Really surprised it changed it to that format with single quotes!

zestyping: No, pyupgrade didn't change it to the single-quote format.  I had previously written the formatting with `%` because it's guaranteed to work everywhere.  Then pyupgrade replaced it with the less-used and unwieldy `.format()` method.  So, in an attempt to rewrite it in a modern way that pyupgrade would like, I then wrote it the Python 3.12 way.

I'm skeptical that pyupgrade is providing value here; it's soaking up developer time on something that shouldn't really matter.

I had a closer look at the pyupgrade repo, it looks like:
  - It doesn't claim to respect `requires-python`, and I don't see any mention of `requires-python` in the code
  - It has command-line options for selecting a Python version, but there's no documentation on how it decides what Python version to target by default
  - The README doesn't explain the command-line options
  - The author seems to actively resist [suggestions](https://github.com/asottile/pyupgrade/issues/1030) to [improve](https://github.com/asottile/pyupgrade/issues/944) the README, and has even rejected many PRs offering improvements to the docs ([643](https://github.com/asottile/pyupgrade/pull/643), [790](https://github.com/asottile/pyupgrade/pull/790), [274](https://github.com/asottile/pyupgrade/pull/274), [275](https://github.com/asottile/pyupgrade/pull/275)), dare I say in a somewhat hostile way

I think it's a liability to make our project depend on something that's not committed to having easy-to-understand behaviour regarding Python versions.  We could invest more work to engineer around that, but I don't see why that would be worth the ongoing maintenance cost — I guess I'm just not seeing a strong case for how it benefits us.  Would you mind if I removed pyupgrade from our precommit hooks?

spwoodcock: You could be right! We use it in a few other repos in the hotosm org - could re-consider it's usage across all of them.

Happy for your to remove from this repo though - thanks for checking in detail 🙏

spwoodcock: Also, looks like ruff can replace the functionality now, under a single tool: https://docs.astral.sh/ruff/rules/#pyupgrade-up (if we want to enable that ruleset)

zestyping: That looks like the right way to go, yeah.  Fewer dependencies, and `ruff` is faster to boot.

In the meantime, the quoting on this line is fixed.

PR #820, opened 2026-05-23
Update split_by_square dependency from fmtm-splitter to area-splitter

### LINKED ISSUE #759: Integrate the Field-TM `area-splitter` module for the split-by-square functionality
- We use shapely to do a custom split by square.
- Note I think we account for no-fly zones, so probably need to verify this will work...
- But the [area-splitter](https://pypi.org/project/area-splitter) module already includes good code (using PostGIS) to split an AOI into tasks.
- Good to centralise around a single implementation, especially if it uses PostGIS instead of shapely.

kygia:

- [X] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation
- [ ] 🧑‍💻 Refactor
- [ ] ✅ Test
- [ ] 🤖 Build or CI

## Related Issue

Fixes #759 

## Describe this PR

Replaces fmtm-splitter with area-splitter, uses PostGIS SQL and requires a DB connection instead of pure python. Updated call site and mocked _area_split_by_square in tests accordingly.

## AI Tool Usage

- [ ] No AI tools were used
- [X] AI tools were used (complete below)

**If AI-assisted:**

- Tool(s) used: Opus 4.6
- What was generated: dependency update suggestions, initial scaffolding for test mocks
- What you reviewed and changed:verified call signatures, confirmed db requirements, confirmed test coverage

## Screenshots

If applicable.

Did you consider other approaches? Why this one?

## Review Guide

Core change is in the split_by_square call site, verify the psycopg connection is passed correctly. Test mock _area_split_by_square to avoid a live DB dependency

## Checklist

- [X] I have read the [Contributing Guide](https://github.com/hotosm/drone-tm/blob/dev/CONTRIBUTING.md)
- [X] I have read the [Code of Conduct](https://docs.hotosm.org/code-of-conduct)
- [X] PR is focused and small
- [X] Tests are included or updated
- [X] I understand all code in this PR and can answer questions about it
- [X] No secrets, credentials, or sensitive data are included
- [X] Commit messages are descriptive
- [X] Related docs and screenshots are updated


kygia:

spwoodcock (changes requested):
Hi! Thanks so much for taking the time 😃

The objective of replacing the current splitter was to remove the shapely dependency and rely on the PostGIS implementation.

Without shapely being replaced there is very little value to this change.

Do you have any thoughts on how the no-fly zone logic could be replaced at the same time? (we need to handle splitting by square, but with the no fly zones omitted).

Thanks again 🙏

spwoodcock:
Also, sorry, I realised I commented a month ago that adding no fly zone handling may be out of scope.

But looking into the changes in the PR and actually seeing the code, it's become clearer to me that without handling this, then the change is probably actually a regression rather than an improvement (increased complexity with little gain) 😄

PR #793, opened 2026-04-18
Create a file transfer util using Flutter for waypoint mission file copy

spwoodcock:
## What type of PR is this? (check all applicable)

- [x] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation
- [ ] 🧑‍💻 Refactor
- [ ] ✅ Test
- [ ] 🤖 Build or CI
- [ ] ❓ Other (please specify)

## Related Issue

Fixes #615
Fixes #769

## Describe this PR

- See plan.md in this PR for full context.
- This is a prototype / demo to test out the functionality proposed.
- The current mobile app isn't working well for people - I wonder if this improves things.
- The context for Flutter is explained in the issue: would be great to have Android / iOS working for this.
- I haven't used Flutter much, hence just letting an LLM just give this a crack first...

Test APK here:
[dronetm-transfer.zip](https://github.com/user-attachments/files/26861275/dronetm-transfer.zip)

## AI Tool Usage

- [ ] No AI tools were used
- [x] AI tools were used (complete below)

**If AI-assisted:**

- Tool(s) used: Opus 4.6 LLM
- What was generated: all of the code in this PR - it's a test / prototype
- What you reviewed and changed: Added the build script after

## Review Guide

- Run `build.sh` to reproduce the APK.
- Install and test it!
- I don't have an RC2 controller to test this unfortunately 😢 

## Alternatives considered

- Fixing the dronetm-mobile app for people.
- The `plan.md` file here has a good analysis of what might be going wrong for them.

## Checklist

- [x] I have read the [Contributing Guide](https://github.com/hotosm/drone-tm/blob/dev/CONTRIBUTING.md)
- [x] I have read the [Code of Conduct](https://docs.hotosm.org/code-of-conduct)
- [x] PR is focused and small
- [x] Tests are included or updated
- [x] I understand all code in this PR and can answer questions about it
- [x] No secrets, credentials, or sensitive data are included
- [x] Commit messages are descriptive
- [x] Related docs and screenshots are updated

PR #823, opened 2026-06-02
feat(transfer-util): restore missing Flutter lib/ Dart sources

FeezyHendrix:
## What this PR does

Restores the **Flutter `lib/` Dart sources** for the `transfer-util` app that are missing from #793.

The root `.gitignore` has a Python-packaging rule (`lib/`) that silently matched and excluded `src/transfer-util/lib/`, so the entire Dart app layer never got committed on `feat/file-transfer-util`. As a result the app cannot build (`flutter build apk` has no entry point) and `test/widget_test.dart` (which imports `package:dronetm_transfer/main.dart`) cannot compile.

This PR targets `feat/file-transfer-util` so it can be merged into #793.

## Changes

- **`build`**: anchored `.gitignore` negation (`!/src/transfer-util/lib/`) so the Dart sources are tracked without affecting the Python `lib/` ignore.
- **`feat`**: the `lib/` layer, written against the **existing Kotlin platform channels** (`MtpTransferPlugin.kt`, `SafTransferPlugin.kt`) — channel names, method args, return shapes, event types and error codes all match the native contract exactly:
  - `models/` — `Mission`, `KmzFile` (with uuid inference), `UsbDeviceInfo`
  - `platform/` — typed `MtpChannel` (method + event channel) and `SafChannel`
  - `services/` — `TransferStrategy` interface + friendly `TransferException`, `MtpTransferStrategy` (direct USB, incl. the async USB-permission handshake), `SafTransferStrategy` (persisted-permission fallback)
  - `state/` — `TransferController` orchestration (auto strategy selection, share/deeplink intake, graceful MTP→SAF fallback, auto-retry/disconnect handling)
  - UI — Material 3 `theme`, phase-driven `HomeScreen`, `MissionSelector` widget, `main` entry point

## Behaviour notes

- Honors DJI's constraint: missions can only **replace** an existing `<uuid>.kmz` in a slot DJI Fly already created — empty list shows explicit "create a mission in DJI Fly first" guidance.
- Seamless flow: auto-detects USB to pick the default strategy, auto-selects the slot matching the file name, one-time SAF folder grant, clear per-phase UI with accessible semantics.
- No HTTP strategies (3 & 4 in `plan.md`) — there is no native counterpart, so they're intentionally out of scope here.

## Verification

Run on Flutter `3.44.1` / Dart `3.12.1`:

- `flutter analyze` → **No issues found!**
- `flutter test` → **All tests passed!** (`App launches`)

> Note: not a full `flutter build apk` yet (needs Android licenses accepted in the build env). The `build.sh` Docker path remains the canonical APK build.

Assisted-by: Claude (opencode)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -89,3 +89,8 @@ src/backend/frontend_html
 *.json
 !**/package.json
 !**/tsconfig.json
+
+# The Python "lib/" packaging ignore above also swallows the Flutter transfer
+# util's Dart sources. Re-include them (anchored so it only affects this app).
+!/src/transfer-util/lib/
+!/src/transfer-util/lib/**
--- a/src/transfer-util/lib/main.dart
+++ b/src/transfer-util/lib/main.dart
@@ -1,0 +1,31 @@
+import 'package:flutter/material.dart';
+import 'package:provider/provider.dart';
+
+import 'screens/home_screen.dart';
+import 'state/transfer_controller.dart';
+import 'theme.dart';
+
+void main() {
+  runApp(const DroneTMTransferApp());
+}
+
+/// Root widget. Owns the single [TransferController] for the whole app.
+class DroneTMTransferApp extends StatelessWidget {
+  const DroneTMTransferApp({super.key});
+
+  @override
+  Widget build(BuildContext context) {
+    return ChangeNotifierProvider<TransferController>(
+      // Build the controller and start listening for launch intents / events
+      // after the first frame, so a missing plugin can never block startup.
+      create: (_) => TransferController()..init(),
+      child: MaterialApp(
+        title: 'DroneTM Transfer',
+        debugShowCheckedModeBanner: false,
+        theme: AppTheme.light(),
+        darkTheme: AppTheme.dark(),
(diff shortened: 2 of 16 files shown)


spwoodcock:
@FeezyHendrix do you have a drone controller for testing by the way?

I have a Potensic Atom 2 controller with me, so can test if the option to select the drone model is added (replicating the existing code we have for the WebUSB implementation in the frontend for both DJI + Potensic).
(only DJI Mini 4/5 Pro and Potensic Atom 2 are needed btw)

Happy to test this if you make progress by Saturday - we will be in the field on Sunday! 🙏 
(this is a huge bonus if it works, but the old cumbersome method works too, so don't worry if it doesn't work out)

FeezyHendrix:
I currently don't have a drone controller, But I would let you know when to test this @spwoodcock

FeezyHendrix:
@spwoodcock I think this can be tested now.

spwoodcock:
Thanks! I'll merge it to the other PR & we can test it from that 🙌

PR #824, opened 2026-06-03
fix(backend): replace remaining hardcoded /tmp paths with tempfile

m4cd4r4:

- [x] 🧑‍💻 Refactor

## Related Issue

Follow-up to #775.

## Describe this PR

When #775 was closed, most of the `/tmp` work had already landed on `dev`, but a handful of hardcoded `/tmp/` paths were still left in the backend. @spwoodcock welcomed a smaller follow-up, so this clears the remaining ones.

Each path now resolves through the `tempfile` module instead of assuming a POSIX `/tmp`, matching the idioms already used elsewhere in the repo:

- `gcp/gcp_routes.py` - GCP upload now goes through a `tempfile.NamedTemporaryFile` (same pattern as `images/image_logic.py`). As a side effect the scratch file is also cleaned up after the S3 upload instead of being left behind.
- `jaxa/upload_dem.py` - DEM processing dir moved to `tempfile.gettempdir()`, keeping the existing per-project subdirectory so the k8s conflict-avoidance comment still holds.
- `projects/classification_routes.py` - reflight plan path via `tempfile.gettempdir()`; existing `background_tasks` cleanup is unchanged.
- `projects/project_logic.py` - three paths across the two terrain-follow blocks. `process_waypoints_and_waylines` now uses `tempfile.mkdtemp()` (which makes the redundant `os.makedirs` unnecessary); `process_task_metrics` keeps its existing behaviour, just portable.
- `waypoints/waypoint_routes.py` - the `generate_wmpl_kmz` and `generate_kmz_with_placemarks` endpoints. The uploaded DEM now lands in a `tempfile.mkstemp` file (matching the existing terrain-follow handler above it in the same file) rather than `f"/tmp/{dem.filename}"`, which also removes a path built straight from a user-supplied filename.

The `drone-flightplan` output writers (`dji.py`, `litchi.py`, etc.) still carry `/tmp/` *default* arguments, but those belong to the vendored package and are better handled in a dedicated upstream PR, so they are intentionally out of scope here.

## AI Tool Usage

- [ ] No AI tools were used

**If AI-assisted:**

- Tool(s) used: Claude Code
- What was generated: the initial edits replacing the hardcoded paths and this description draft.
- What you reviewed and changed: I located each remaining `/tmp` path, chose the matching tempfile idiom per call site, kept `process_task_metrics` behaviour-preserving rather than "fixing" a long-dormant code path I could not verify, confirmed the imports, and ran py_compile + ruff locally. I understand and stand behind every line.

For the GCP upload I considered a like-for-like `mkstemp` swap, but the upload there never deleted its scratch file, so `NamedTemporaryFile` as a context manager is both portable and tidier. For `jaxa/upload_dem.py` I kept the stable per-project subdirectory rather than switching to a random `mkdtemp`, to preserve the documented k8s conflict-avoidance intent.

## Review Guide

Behaviour is unchanged except for the two noted side effects (GCP scratch file is now cleaned up; the DEM upload filename is no longer used to build a path). `python -m py_compile` and `ruff check` pass on the five touched files.

## Checklist

- [x] PR is focused and small

### DIFF
--- a/src/backend/app/gcp/gcp_routes.py
+++ b/src/backend/app/gcp/gcp_routes.py
@@ -1,3 +1,4 @@
+import tempfile
 import uuid
 from typing import Annotated, List
 
@@ -57,12 +58,12 @@ async def save_gcp_file(
     The file is stored at `projects/{project_id}/gcp.txt` in S3
     and will be automatically included when final processing is triggered.
     """
-    gcp_file_path = f"/tmp/{uuid.uuid4()}"
-    with open(gcp_file_path, "wb") as f:
-        f.write(await gcp_file.read())
+    with tempfile.NamedTemporaryFile(suffix=".txt") as gcp_file_tmp:
+        gcp_file_tmp.write(await gcp_file.read())
+        gcp_file_tmp.flush()
 
-    s3_path = f"projects/{project.id}/gcp.txt"
-    add_file_to_bucket(settings.S3_BUCKET_NAME, gcp_file_path, s3_path)
+        s3_path = f"projects/{project.id}/gcp.txt"
+        add_file_to_bucket(settings.S3_BUCKET_NAME, gcp_file_tmp.name, s3_path)
     log.info(f"GCP file saved for project {project.id} by user {user_data.id}")
 
     return {"message": "GCP file saved successfully", "project_id": str(project.id)}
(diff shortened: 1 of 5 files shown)


m4cd4r4:

spwoodcock (approved):
Looks good to me - thanks for the PR @m4cd4r4! 🙏

spwoodcock:
@all-contributors add @m4cd4r4 for code

spwoodcock:
Ah I forgot we aren't using all-contributors in this repo! 😁

PR #831, opened 2026-06-09
feat(transfer-util): field diagnostics for flightplan transfer

FeezyHendrix:
## Why

Flightplan transfer is DroneTM's #1 field issue. The transfer app works, but when it fails on a phone it surfaces a single generic error with no field-collectable record of *why* — so "works on 1/3 phones" is unactionable. This adds a diagnostics layer that turns each attempt (or a standalone scan) into a structured, exportable report pinpointing the failing step and the raw USB/MTP facts underneath it.

## What's in it

**Native capture** (`MtpTransferPlugin`)
- `getEnvironment` — phone make/model, Android SDK, USB-host feature, app version
- `dumpUsbDevices` — every USB device *unfiltered*, with full interface detail, so an RC2 enumerating with an unexpected product id is visible instead of looking identical to "nothing plugged in"
- `dumpMtpTree` — storage volumes + the DJI-path traversal (deepest segment reached, the missing segment, siblings present where it stopped). Runs on a dedicated worker thread (no UI block / ANR), bounded by depth/width/node budget, uses `compressedSizeLong` for >2 GiB media.

**Always-on step log** (MTP + SAF flows)
- Per-step ok / fail / **timeout**, distinguishing a permission denial from a missing permission broadcast (the Android 14 `RECEIVER_EXPORTED` class of hang)
- Captures the previously-discarded `MtpDeviceInfo`

**Export** (`DiagnosticsPlugin` + Dart) — all offline-first, capture now / send later:
- Share sheet (FileProvider), clipboard, save-to-Downloads (MediaStore on API 29+, app-specific external dir pre-Q to avoid a silent permission failure), and an offline-queued HTTP POST that flushes when a configurable endpoint is reachable

**Diagnostics screen** (troubleshoot icon, top-right): one-tap **Run full diagnostic scan** that needs no KMZ, a live colour-coded log, an export bar, and endpoint config.

No new pub dependencies. Adds the `INTERNET` permission and a `FileProvider`.

## Verification

- `flutter analyze` clean; `flutter build apk --debug` succeeds.
- Reviewed across Kotlin/MTP-API correctness, Dart correctness, manifest/FileProvider, and the offline queue. The findings are folded in: off-thread tree walk, `compressedSizeLong`, run-scoped summary verdict, pre-Q save without a runtime permission, SAF per-step logging, and cache reset on "Clear log".

## How to test (DJI RC2)

1. `cd src/transfer-util && flutter install` (or `adb install -r build/app/outputs/flutter-apk/app-debug.apk`)
2. Plug the phone into the RC2 with a **data** cable → open the app → troubleshoot icon → **Run full diagnostic scan**
3. **Share** the report. Repeat on each test phone — the reports show exactly why the failing phones fail (no-OTG vs charge-only cable vs permission-timeout vs PID-mismatch vs no waypoint slot).

## Notes

- Base is `transfer-util-lib` so the diff is diagnostics-only; retarget to `feat/file-transfer-util` / `dev` if preferred.
- `gradle.properties` carries the Flutter 3.44 migrator's `builtInKotlin` / `newDsl=false` flags, needed for the build on that SDK.
- Design doc: `src/transfer-util/docs/plans/2026-06-08-mtp-transfer-diagnostics-design.md`

### DIFF
--- a/src/transfer-util/android/app/src/main/AndroidManifest.xml
+++ b/src/transfer-util/android/app/src/main/AndroidManifest.xml
@@ -8,6 +8,9 @@
     <uses-permission android:name="android.permission.WRITE_EXTERNAL_STORAGE" />
     <uses-permission android:name="android.permission.MANAGE_EXTERNAL_STORAGE" />
 
+    <!-- Uploading queued diagnostic reports when connectivity returns -->
+    <uses-permission android:name="android.permission.INTERNET" />
+
     <application
         android:label="DroneTM Transfer"
         android:name="${applicationName}"
@@ -72,6 +75,17 @@
         <meta-data
             android:name="flutterEmbedding"
             android:value="2" />
+
+        <!-- Shares diagnostic reports out of the app's cache via the share sheet -->
+        <provider
+            android:name="androidx.core.content.FileProvider"
+            android:authorities="${applicationId}.fileprovider"
+            android:exported="false"
+            android:grantUriPermissions="true">
+            <meta-data
+                android:name="android.support.FILE_PROVIDER_PATHS"
+                android:resource="@xml/file_paths" />
+        </provider>
     </application>
 
     <queries>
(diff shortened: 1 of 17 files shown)


spwoodcock:
Thanks!! I'll merge this in, as it will definitely be helpful to verify how reliable the MTP is afterwards.

But I have a feeling this problem is deeper and needs a serious look from a low-level Android developer or something.
Gen-AI probably won't help in this case, but I'm willing to be proven wrong! 😄

PR #850, opened 2026-07-16
Image footprint coverage

### LINKED ISSUE #819: Image coverage estimation is too generous, giving a false impression of completion
Image

- This means the 'Identify Flight Gaps' button may not work (need to verify).
- The user should have an accurate representation of how much area if covered, ideally we could show some hairline squares of the image coverage projected onto the map, so the user can actually see what the photos cover (although with lots of overlap, this might look messy).

shreya-hegde:

- [ ] 🍕 Feature
- [x] 🐛 Bug Fix
- [ ] 📝 Documentation
- [ ] 🧑‍💻 Refactor
- [x] ✅ Test
- [ ] 🤖 Build or CI

## Related Issue

Fixes #819

## What changed

- Added rectangular image footprints to task verification.
- Coverage now uses the union of those footprints instead of over-generous point coverage.
- Draws footprint outlines on the map.
- Clicking an image highlights its matching footprint.
- Added FC9313 for DJI Mini 5 Pro.

## Notes

- Footprints use project altitude first, then EXIF relative altitude.
- Avoided EXIF absolute altitude because it is height above sea level and made coverage too large.

## Tested

- Tested locally with Mt Agung data.
- Task #155 had 10 assigned images.
- Coverage changed from 100% to 14%.
- 
## AI Tool Usage

- [ ] No AI tools were used
- [ x] AI tools were used (complete below)

**If AI-assisted:**

- Tool(s) used: Codex
- What was generated: Draft code changes, explanations, and debugging guidance.
- What you reviewed and changed: Reviewed the code, ran local/Docker tests, checked UI behavior, and adjusted altitude/footprint logic.

## Screenshots

If applicable.

Did you consider other approaches? Why this one?

## Review Guide

Notes for the reviewer. How to test this change?

Main things to look at:

- image_footprints.py: footprint size/shape calculation and union coverage logic.
- image_classification.py: task verification now uses footprint coverage instead of the old circular buffer coverage.
- TaskVerificationModal.tsx: footprint outlines are drawn on the map and selected image footprints are highlighted.

Things I’d especially like feedback on:

- Whether using project altitude first, then EXIF relative altitude, is the right assumption.
- Whether the Identify Flight Gaps flow should be updated to use the same footprints.

## Checklist

- [ x] I have read the [Contributing Guide](https://github.com/hotosm/drone-tm/blob/dev/CONTRIBUTING.md)
- [ x] I have read the [Code of Conduct](https://docs.hotosm.org/code-of-conduct)
- [x ] PR is focused and small
- [ x] Tests are included or updated
- [ x] I understand all code in this PR and can answer questions about it
- [ x] No secrets, credentials, or sensitive data are included
- [x ] Commit messages are descriptive
- [ ] Related docs and screenshots are updated

## [optional] What gif best describes this PR or how it makes you feel?
https://giphy.com/gifs/square-triangle-shape-slap-V6kvt4UVqfHha3Eb19 (cuz the rectangles/squares are the superior shapes in image coverage)

### DIFF
--- a/src/backend/app/images/image_footprints.py
+++ b/src/backend/app/images/image_footprints.py
@@ -1,0 +1,155 @@
+import math
+
+import pyproj
+from shapely.affinity import rotate
+from shapely.geometry import Polygon, mapping, shape
+from shapely.ops import transform, unary_union
+
+
+# Convert lon/lat to meters for footprint and area math.
+# A degree is not a fixed ground distance, so we avoid area math in EPSG:4326.
+projector = pyproj.Transformer.from_crs("EPSG:4326", "EPSG:3857", always_xy=True)
+inverse_projector = pyproj.Transformer.from_crs(
+    "EPSG:3857", "EPSG:4326", always_xy=True
+)
+
+# Fallbacks when EXIF/project metadata is missing.
+DEFAULT_DIAGONAL_FOV_DEG = 82.1
+DEFAULT_IMAGE_WIDTH = 4000
+DEFAULT_IMAGE_HEIGHT = 3000
+
+
+def _as_float(value) -> float | None:
+    # EXIF values can be messey as strings, blanks, or missing.
+    # This function converts values into numbers or None.
+    if value is None:
+        return None
+
+    try:
+        return float(value)
+    except (TypeError, ValueError):
+        return None
+
+
+def _footprint_size_meters(
+    gsd_cm_px: float | None,
+    altitude_m: float | None,
+    image_width: float | None,
+    image_height: float | None,
+) -> tuple[float, float] | None:
+    """Estimate the ground width/height covered by one image."""
+    # Prefer GSD: it tells us directly how much ground each pixel covers.
+    gsd_cm_px = _as_float(gsd_cm_px)
+    altitude_m = _as_float(altitude_m)
+    image_width = _as_float(image_width) or DEFAULT_IMAGE_WIDTH
+    image_height = _as_float(image_height) or DEFAULT_IMAGE_HEIGHT
+
+    if gsd_cm_px and gsd_cm_px > 0:
+        # Example: 2 cm/px * 4000 px / 100 = 80m on the ground.
+        return (
+            gsd_cm_px * image_width / 100,
+            gsd_cm_px * image_height / 100,
+        )
+
+    # If no GSD: altitude + camera FOV gives an approximate ground rectangle.
+    if altitude_m and altitude_m > 0:
+        # Keep the rectangle shaped like the image, usually 4:3.
+        aspect_ratio = image_width / image_height
+        diagonal_m = 2 * altitude_m * math.tan(
+            math.radians(DEFAULT_DIAGONAL_FOV_DEG) / 2
+        )
+        # Convert the estimated diagonal into width and height.
+        height_m = diagonal_m / math.sqrt(1 + aspect_ratio**2)
+        width_m = aspect_ratio * height_m
+        return width_m, height_m
+
+    return None
+
+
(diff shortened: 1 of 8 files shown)


shreya-hegde:

src/backend/app/images/image_footprints.py
```diff
@@ -0,0 +1,155 @@
...
+        image.get("image_width"),
+        image.get("image_height"),
+    )
+    if not footprint_size:
+        return None
+
+    # Build in meters first, then convert back to lon/lat only for display.
+    width_m, height_m = footprint_size
+    # Convert image GPS point into meter coordinates
+    center = transform(projector.transform, shape(image["location"]))
```
spwoodcock, line 89: This will be an EPSG:3857 (web mercator) coordination, but the footprint calculated below is true ground meters, so the calculations as we get closer to the equator will be increasingly incorrect.

I would propose a simple solution is to correct by the `1/cos(latitude)` adjustment that web mercator uses:

```python
center = transform(projector.transform, shape(image["location"]))
x, y = center.x, center.y

# EPSG:3857 stretches ground distance by 1/cos(lat). Scale the rectangle
# so its footprint is stretched to match the mercator projection.
lat_deg = image["location"]["coordinates"][1]
mercator_scale = 1 / math.cos(math.radians(lat_deg))
half_width = (width_m / 2) * mercator_scale
half_height = (height_m / 2) * mercator_scale
```

src/backend/tests/test_image_footprints.py
```diff
@@ -0,0 +1,54 @@
...
+    image_footprints_feature_collection,
+    inverse_projector,
+)
+
+
+def test_image_footprint_feature_collection_contains_polygon():
+    # A tiny fake image where GSD makes the footprint exactly 10m x 10m.
+    image = {
+        "id": "image-1",
+        "location": {"type": "Point", "coordinates": [0, 0]},
```
spwoodcock, line 16: Both these tests uses a 0,0 coord, so won't be affected by the web meracator stretching that I mentioned.

It would be good to add an additional test at say 45 latitude, to ensure our fix handles it well.

(perhaps add the tests first, run the test and see it fail, apply the fix, run again and see it pass 😄 )

src/backend/app/images/image_footprints.py
```diff
@@ -0,0 +1,155 @@
...
+                "type": "Feature",
+                "geometry": mapping(footprint_wgs84),
+                "properties": {"image_id": str(image["id"])},
+            }
+        )
+
+    return {"type": "FeatureCollection", "features": features}
+
+
+def coverage_percentage_from_footprints(
```
spwoodcock, line 136: This is great 👍 

We calculate coverage in this endpoint, but also for the entire project in `get_project_coverage`.

It would be great to refactor the usage there to also use this updated code, so the project level coverage is calculated correctly (merging this as-is would cause two different values between the two implementations. The old implementation can be removed / cleaned up)

src/backend/app/images/image_footprints.py
```diff
@@ -0,0 +1,155 @@
...
+
+    return footprint
+
+
+def image_footprints_feature_collection(images: list[dict]) -> dict:
+    """Return map-ready GeoJSON outlines for all image footprints."""
+    # Frontend map layers need GeoJSON in lon/lat.
+    features = []
+    for image in images:
+        footprint = image_footprint_polygon(image)
```
spwoodcock, line 119: **any way we could consolidate these two footprint calcs, so we only have to do it once per image?

spwoodcock (changes requested):
Really nice work!! This is looking great 😄 

Left some small comments. The main one is about the mercator project towards the poles, which will distort the footprint calcs a lot

PR #868, opened 2026-08-16
fix(backend): use tempfile for flightplan output instead of /tmp

### LINKED ISSUE #597: Replace usage of /tmp directory with Python tempfile module

StepFPV:

- [ ] 🍕 Feature
- [x] 🐛 Bug Fix
- [ ] 📝 Documentation
- [x] 🧑‍💻 Refactor
- [x] ✅ Test
- [ ] 🤖 Build or CI

## Related Issue

Fixes #597

## Describe this PR

The output writers defaulted to hardcoded paths under `/tmp`: `/tmp/mission.csv`, `/tmp/mission.waypoints`, `/tmp/mission.plan` and `/tmp/` for DJI. That causes two distinct problems on a deployed server:

1. **Collisions.** Those are fixed, global filenames. Two users generating a flightplan at the same time write to the same file, and one overwrites the other.
2. **Accumulation.** Nothing ever removes the files, which is the disk-filling behaviour described in the issue.

The DJI writer was the worst of the two, because it doesn't just leave the final file behind. Building a `.kmz` means creating a `flight/` directory, writing `waylines.wpml` into it, copying that same file into a `wpmz/` subdirectory, and zipping. All of that scratch state was left on disk.

Measured with a throwaway script, pointing the writers at a sandbox directory and generating three flightplans:

```
before                                after
  mission.csv                           mission.csv
  flight/output.kmz                     output.kmz
  flight/waylines.wpml
  flight/wpmz/waylines.wpml
```

Roughly 24 KB of scratch files per DJI flightplan, permanently. And `mission.csv` is a single file for all three runs, which is the collision problem made visible.

### Changes

**`drone_flightplan/output/output_paths.py`** (new) — a `resolve_output_path()` helper. `None` means "give me a managed temporary path"; an explicit path is honoured, with missing parent directories created.

**`litchi.py`, `mavlink.py`, `qgroundcontrol.py`** — `output_file_path` defaults to `None` instead of a `/tmp` path. Resolution happens immediately before writing, so a validation failure (e.g. empty FeatureCollection) doesn't leave an empty temp file behind.

**`dji.py`** — the `waylines.wpml` / `wpmz` tree is now built inside a `tempfile.TemporaryDirectory()`, so only the requested `.kmz` survives the call. `create_zip_file()` takes an optional destination so the archive can land outside that scratch directory; omitting it keeps the previous behaviour.

**`app/waypoints/flightplan_output.py`** — `build_flightplan_download_response()` accepts `cleanup_dir` and attaches a `BackgroundTask` that removes it. The file can't be deleted when the handler returns, because `FileResponse` streams it afterwards; a background task runs at the right moment, once the response has been sent.

**`app/waypoints/waypoint_routes.py`** — all three flightplan endpoints generate into a per-request temp directory, cleaned up after the response. If generation raises there is no response and therefore no background task, so those paths remove the directory directly.

The uploaded DEM in `generate_wmpl_kmz` is also removed in a `finally` block; previously it stayed on disk. While adding that I noticed `dem=dem_path if dem else None` referenced `dem_path`, which is only assigned inside `if terrain_follow:`. Uploading a DEM without `terrain_follow` raised `NameError` and returned a 500. Initialising `dem_path = None` up front was needed for the `finally` anyway and fixes that as a side effect. Happy to split that out if you'd rather keep it separate.

**`app/projects/project_logic.py`** — terrain-follow generation used `tempfile.gettempdir() / "output_file_with_elevation.geojson"`, one fixed filename shared by every concurrent call, and a DEM directory that was never removed. Both are now inside a scoped `TemporaryDirectory`. This is the same class of bug rather than the literal `/tmp` string, so say the word if you'd prefer it in its own PR.

### Behaviour change worth flagging

Callers that relied on the default output location will now get a unique temporary path instead of a fixed one, and must use the returned path. The return value has always been there, and `drone-flightplan` is published on PyPI, so I kept one compatibility affordance: `create_wpml()` still accepts an existing directory as `output_file_path` and writes `output.kmz` inside it.

That also unifies the two conventions noted in the existing comment in `waypoint_routes.py` — DJI treated the argument as a directory while every other writer treated it as a file path. It's now consistently a file path, with the directory form still working.

## AI Tool Usage

- [ ] No AI tools were used

**If AI-assisted:**

- Tool(s) used: Claude Code (`Assisted-by` trailer is in the commit)
- What was generated: the implementation and the tests, from my analysis of the issue and the surrounding code.
- What you reviewed and changed: I walked through the whole diff before opening this, and I understand and stand behind every change in it. The scope was deliberately widened during review once it became clear the hardcoded defaults were only part of the leak — the DJI scratch tree and the missing cleanup in the routes are where the disk actually fills up. I also decided to keep the directory-argument compatibility path in `create_wpml()` rather than making a clean break, since the package is published separately. The two questions I raised on the issue are still open, and I'm happy to reshape the public API if you prefer a different direction.

## Screenshots

Not applicable — the before/after evidence is the directory listing above.

**Encapsulating temp handling fully inside the writers** (a `TemporaryDirectory` around the whole function). Rejected: these functions return a path the caller then serves to the user, so a context manager would delete the file before it could be read. `mkstemp` is used instead, with ownership documented as passing to the caller. The context manager is used only where the file's life genuinely ends inside the function — the DJI scratch tree.

**Leaving the writers alone and only cleaning up in the routes.** That would fix the disk usage but not the collisions between concurrent missions, and would leave the misleading `/tmp` defaults in a published package.

**Making `output_file_path` a required argument.** Cleaner in isolation, but a breaking change for every external caller with no compatibility path, for little benefit.

## Review Guide

Suggested reading order: `output_paths.py` (the helper), then `dji.py` (the only non-trivial writer change), then `flightplan_output.py` and `waypoint_routes.py` (cleanup after the response is sent).

New tests are in `src/backend/tests/test_flightplan_output_paths.py`. The `isolated_tmpdir` fixture repoints `tempfile.tempdir` at an empty directory, which is what makes "exactly one file remains" assertable.

To confirm they're meaningful rather than tautological: stash the source changes, keep the test file, and the suite reports **15 failed, 2 passed**. With the fix applied, **17 passed**.

Manual check: generate a flightplan download for a DJI project and for a non-DJI drone type, confirm the file downloads intact and that no `flightplan_*` directory is left in the container's temp dir afterwards.

**What I could not verify:** the full backend suite. It needs the Docker Compose stack (Postgres, Dragonfly, S3) and I couldn't run it on this machine. The 17 new tests were run in isolation against the package and `flightplan_output.py`, with `osgeo` stubbed since GDAL isn't installable here. `ruff check` and `ruff format --check` are clean across `src/backend`. Everything else is down to CI.

## Checklist

- [x] PR is focused and small
- [ ] Related docs and screenshots are updated

## [optional] What gif best describes this PR or how it makes you feel?

Someone quietly emptying the wastepaper basket that everyone had been stepping around for two years.

### DIFF
--- a/src/backend/app/projects/project_logic.py
+++ b/src/backend/app/projects/project_logic.py
@@ -244,22 +244,25 @@ async def process_task_metrics(db, tasks_data, project):
         }
 
         if project.is_terrain_follow:
-            dem_path = os.path.join(tempfile.gettempdir(), str(uuid.uuid4()), "dem.tif")
             points = create_waypoint(**waypoint_params)
-            try:
-                get_file_from_bucket(
-                    settings.S3_BUCKET_NAME,
-                    f"projects/{project.id}/dem.tif",
-                    dem_path,
-                )
+            # Scoped temp dir: the DEM directory was never removed, and the
+            # elevation file used a fixed name shared by every concurrent call
+            with tempfile.TemporaryDirectory(prefix="dem_") as temp_dir:
+                dem_path = os.path.join(temp_dir, "dem.tif")
                 outfile_with_elevation = os.path.join(
-                    tempfile.gettempdir(), "output_file_with_elevation.geojson"
+                    temp_dir, "output_file_with_elevation.geojson"
                 )
-                add_elevation_from_dem(dem_path, points, outfile_with_elevation)
-                with open(outfile_with_elevation) as inpointsfile:
-                    points_with_elevation = inpointsfile.read()
-            except Exception:
-                points_with_elevation = points
+                try:
+                    get_file_from_bucket(
+                        settings.S3_BUCKET_NAME,
+                        f"projects/{project.id}/dem.tif",
+                        dem_path,
+                    )
+                    add_elevation_from_dem(dem_path, points, outfile_with_elevation)
+                    with open(outfile_with_elevation) as inpointsfile:
+                        points_with_elevation = inpointsfile.read()
+                except Exception:
+                    points_with_elevation = points
 
             if (
                 isinstance(points_with_elevation, dict)
--- a/src/backend/app/waypoints/flightplan_output.py
+++ b/src/backend/app/waypoints/flightplan_output.py
@@ -1,6 +1,9 @@
+import shutil
+
 from drone_flightplan.drone_type import DRONE_PARAMS, DroneType
 from fastapi import HTTPException
 from fastapi.responses import FileResponse
+from starlette.background import BackgroundTask
 
 FLIGHTPLAN_OUTPUTS = {
     "DJI_WMPL": {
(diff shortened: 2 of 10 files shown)


StepFPV:

spwoodcock:
Really essential work!

Have you ran a full E2E test after these changes?

1. Run DroneTM locally
2. Create a project with terrain following
3. Generate flightplans, and inspect the whole process succeeded
4. (optional, but preferred): put the flightplan into a drone to actually confirm it's fine. Not essential, as you can do a fair amount by just inspecting the geojson output / kmz directly instead

PR #872, opened 2026-08-22
fix: correct axis order in DEM elevation sampling

### LINKED ISSUE #718: Update elevation extraction logic to first extract the data axis to CRS axis mapping from DEMs
Feedback from https://github.com/hotosm/drone-tm/pull/590

- @andrewharvey reported that the X/Y axis were flipped in their output running `drone-flightplan` manually, with a custom DEM.
- Originally we thought it was an issue with GDAL after version upgrades, but we have since also upgraded to Debian Trixie (13) and GDAL 3.10.3 without issues.
- The likely problem is in the DEM we are using - the JAXA DEMs probably have a flipped [data axis to CRS axis mapping](DEM files report a different Data axis to CRS axis mapping? In that case we probably need to detect this when reading the DEM file.) (original comment: "DEM files report a different Data axis to CRS axis mapping? In that case we probably need to detect this when reading the DEM file.").
- This is not a problem for the standard workflow using JAXA, but we need to account for the different types of DEMs users may upload themselves.

dulcetberg:
## Summary

Fixes #718.

`add_elevation_from_dem()` (in `drone_flightplan/add_elevation_from_dem.py`) hardcoded a swap of the coordinate transform's output:

```python
pointXYRasterCRS = transform.TransformPoint(geom.GetX(), geom.GetY())
mapX = pointXYRasterCRS[1]
mapY = pointXYRasterCRS[0]
```

This assumes the transform always returns `(lat, lon)`-order results. That only holds for **geographic** CRS rasters (like the app's default WGS84 DEM), where GDAL's default authority-compliant axis order happens to be `(lat, lon)`.

For a **projected** CRS raster (e.g. UTM) — a realistic case for a user-uploaded custom DEM, per the issue's own note about "different types of DEMs users may upload themselves" — authority-compliant order already matches `(x, y)`/`(easting, northing)`. The same hardcoded swap then corrupts an otherwise-correct transform instead of fixing a swapped one, so points get sampled at the wrong pixel (or land outside the raster entirely, silently falling back to elevation `0`).

### Fix

Explicitly set both `SpatialReference` objects (`rasterSR`, `pointSR`) to `OAMS_TRADITIONAL_GIS_ORDER` right after they're built, which always resolves to `(x, y)` regardless of what the underlying CRS's authority declares. This removes the need to guess/hardcode a swap at all — the fix is correct for both geographic and projected DEMs, not just the default case.

## Test plan

- [x] Added `tests/test_add_elevation_from_dem.py` (new — this package had no test coverage before). Builds a synthetic UTM-projected raster (the specific case that actually exercises the bug — a plain WGS84 raster does *not*, since the old swap happens to cancel out GDAL's implicit one in that case) and asserts the elevation sampled at a known point matches the known pixel value.
- [x] Verified the new test **fails** against the pre-fix code (`AssertionError: Expected elevation 6.0 ..., got 0.0`) and **passes** against the fix.
- [x] Confirmed the fixed module still imports cleanly.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/backend/packages/drone-flightplan/drone_flightplan/add_elevation_from_dem.py
+++ b/src/backend/packages/drone-flightplan/drone_flightplan/add_elevation_from_dem.py
@@ -56,6 +56,7 @@ def add_elevation_from_dem(raster_file, points, outfile) -> int:
     rasterSR = osr.SpatialReference()
     rasterSR.ImportFromProj4(r.GetProjection())
     log.info(f"\nRaster Coordinate Reference System: \n{rasterSR}")
+    rasterSR.SetAxisMappingStrategy(osr.OAMS_TRADITIONAL_GIS_ORDER)
 
     # Get the raster band (if it's a DEM, this should be the only band)
     # TODO this would be a good time to check that it's a single-band
@@ -80,6 +81,7 @@ def add_elevation_from_dem(raster_file, points, outfile) -> int:
     p = ogr.Open(points)
     lyr = p.GetLayer()
     pointSR = lyr.GetSpatialRef()
+    pointSR.SetAxisMappingStrategy(osr.OAMS_TRADITIONAL_GIS_ORDER)
     pointLD = lyr.GetLayerDefn()
     log.info(f"\nPoint layer Coordinate Reference System: {pointSR.GetName()}")
 
@@ -107,8 +109,8 @@ def add_elevation_from_dem(raster_file, points, outfile) -> int:
     for feature in lyr:
         geom = feature.GetGeometryRef()
         pointXYRasterCRS = transform.TransformPoint(geom.GetX(), geom.GetY())
-        mapX = pointXYRasterCRS[1]
-        mapY = pointXYRasterCRS[0]
+        mapX = pointXYRasterCRS[0]
+        mapY = pointXYRasterCRS[1]
         pixcoords = gdal.ApplyGeoTransform(reverse, mapX, mapY)
         pixX = math.floor(pixcoords[0])
         pixY = math.floor(pixcoords[1])
--- a/src/backend/packages/drone-flightplan/tests/test_add_elevation_from_dem.py
+++ b/src/backend/packages/drone-flightplan/tests/test_add_elevation_from_dem.py
@@ -1,9 +1,8 @@
 import json
 
 import numpy as np
-from osgeo import gdal, osr
-
 from drone_flightplan.add_elevation_from_dem import add_elevation_from_dem
+from osgeo import gdal, osr
 
 # A point in Kathmandu, used as the single test point (WGS84 lon/lat, as a
 # GeoJSON file would normally provide).
--- a/src/backend/packages/drone-flightplan/tests/test_add_elevation_from_dem.py
+++ b/src/backend/packages/drone-flightplan/tests/test_add_elevation_from_dem.py
@@ -1,0 +1,103 @@
+import json
+
+import numpy as np
+from osgeo import gdal, osr
+
+from drone_flightplan.add_elevation_from_dem import add_elevation_from_dem
+
+# A point in Kathmandu, used as the single test point (WGS84 lon/lat, as a
+# GeoJSON file would normally provide).
+POINT_LON, POINT_LAT = 84.5, 27.5
+
+# Raster: 10 columns x 4 rows, UTM zone 44N (EPSG:32644) -- a projected CRS,
+# representative of a "user-uploaded" DEM rather than the app's default
+# geographic (WGS84) DEM. Pixel value = column index, constant down each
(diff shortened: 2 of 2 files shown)


spwoodcock:
Sorry I have to ask: are you an AI?

Multiple PRs across many repos in such a short time. AI generated descriptions.

You can never be sure 😅

dulcetberg:
> Sorry I have to ask: are you an AI?
> 
> Multiple PRs across many repos in such a short time. A generated descriptions 😅

Admittedly I am human but I did use Claude AI to help me with the code, clearly. I’m rather new to this and was exploring my options. I’m very sorry if this is against the rules but I figured it’d be a good way to contribute and learn. I’ve taken a few classes on geospatial python and understand the terminology well enough but I’m not good with code. The AI has helped me understand a bit and I’m still learning.

Sorry for the confusion,
Brian Bergstrom
dulcetberg@gmail.com

spwoodcock:
Thanks Brian - appreciate the honesty!

I'm very happy to welcome new contributors, but there are a few things to consider with AI / LLM usage:
https://responsibleai.guide/

Perhaps more importantly, as a contributor: https://responsibleai.guide/ai-assisted-coding-guide/

Thanks for already disclosing the AI usage in the descriptions - it's really helpful 🙏 

The key thing to consider here is that while generating the code may take your a few minutes, it can take the maintainer multiple hours to review and test each PR thoroughly - a huge time drain.

I would propose:
1. If you don't understand the code, then don't submit it
2. If you have read through all the code and understand it, then great! Feel free to submit, but ensure you have tested the changes yourself thoroughly, and can prove that with some screenshots / logs (this helps the maintainer - in this case me - a lot)

We really appreciate you trying to help out with all these open issues!

dulcetberg:
Thanks for the thoughtful reply — I appreciate you taking the time to lay that out rather than just shutting it down.

I went and actually read through the responsible AI guide you linked, not just skimmed it. A few things stuck with me: the "you are the author, you are accountable" framing, and the point about testing changes yourself rather than just handing over something that runs. That second one especially I realized today that I'd been letting Claude run the verification and report back to me, instead of running it myself and seeing the results firsthand. That's on me to fix, not a small thing. But it was indeed me making all the critical decisions about the code once it was explained.

Going forward: I'm going to slow down, read through everything line by line before it goes out, write my own PR descriptions, and actually test changes myself — I do geospatial Python in Jupyter as part of my background, so that's a natural place for me to verify things. Like you suggested, I'll start including screenshots of that testing on future PRs, so you can see I actually ran and checked the code myself rather than just taking my word for it.

I'm new to open source and clearly still learning the norms here. I really appreciate you being direct with me about it instead of just closing the door. I'd like to keep contributing the right way, as I think this work is really cool.

I'm unemployed and looking for volunteer GIS hours to count towards professional experience in my GISP application. A paying job is not what I am after; what I really need is the experience and feedback from people like you, as I have very little and GIS is my passion. I'm tired of applying to countless jobs and getting crickets, so I thought I would take the risk this way, even if it fails. 

Again, I appreciate your feedback and you not closing the door, and next time will be more thought out. Thank you for your time and look forward to making more contributions where I can.

— Brian

spwoodcock:
Thanks for understanding 😄 

I'm very happy to work with you & help support any code changes you may wish to make.

We work with many junior / new developers at HOT, so as long as the PRs are in good faith and in line with the responsible ai guide, then I'm more than happy to help 👍

spwoodcock:
This could be a great change!

Can you explain a bit more about how it works?

I don't really understand: does it make a guess for the ordering based on coord bounds?

The PR swaps the ordering of the indexed coords, so we need to have 100% certainty in this, as otherwise it would give totally wrong elev values and potentially crash a drone

dulcetberg:
@spwoodcock No guessing, and nothing inspects the coordinate values. Two lines declare the axis order up front:

rasterSR.SetAxisMappingStrategy(osr.OAMS_TRADITIONAL_GIS_ORDER)
pointSR.SetAxisMappingStrategy(osr.OAMS_TRADITIONAL_GIS_ORDER)

Both run before osr.CoordinateTransformation(pointSR, rasterSR) is built, and the transform reads the strategy off those objects at build time. Output is then always (x, y), so the swap gets deleted rather than replaced with something cleverer.

What convinced me: on a projected CRS the new lines change nothing. Both versions give the same transform output on my UTM test raster, (845828.367, 3046698.092), since authority-compliant order there is already easting then northing. Only the swap differs. It feeds easting 3046698 into a raster spanning 839328 to 849328, about 2207 pixels off a 10 pixel wide raster, and the out of bounds branch returns elevation 0 with no error raised. Geographic rasters land on the same pixel either way, so the default DEM is unaffected.

Also fixed a row positioning bug in my own test that I found while checking this. Tested on GDAL 3.12, CI pins 3.10.3.

dulcetberg:

PR #878, opened 2026-08-31
feat: replace Nominatim reverse geocoding with pg-nearest-city

### LINKED ISSUE #554: Replace usage of Nominatim with pg-nearest-city
**Is your feature request related to a problem? Please describe.**

- Nominatim is a great web reverse geocoder, hosted for free by OSM.
- We don't need detailed reverse geocoding to an address. Instead projects can be reverse geocoded to the nearest city of 1000 people.

**Describe the solution you'd like**

- We made a package for FieldTM to do exactly that: github.com/hotosm/pg-nearest-city/
- Using the underlying PostGIS database, we can reverse geocode during project creation, entirely offline.
- Usage of Nominatim can be removed, also reducing the load on the OSM-hosted server during testing / project creation.

Bloom79:

- [x] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation
- [ ] 🧑‍💻 Refactor
- [ ] ✅ Test
- [ ] 🤖 Build or CI

## Related Issue

Fixes #554

## Describe this PR

Removes the dependency on the OSM-hosted Nominatim service for reverse geocoding during project creation, as requested in #554:

- **Backend**: new `GET /api/projects/nearest-city/?lon=&lat=` endpoint (login required) that reverse geocodes the project centroid to `{city, country}` using [pg-nearest-city](https://github.com/hotosm/pg-nearest-city) (v2.0.0) over the existing psycopg PostGIS connection — fully offline. Returns `{city: null, country: null}` when no match, mirroring how Field-TM/Tasking Manager adopted the package.
- **Frontend**: `getCountry()` (axios call to `nominatim.openstreetmap.org/reverse`) is replaced by `getNearestCity()` calling the new endpoint; the create-project flow keeps populating `projectCountry` (used for the drone-altitude lookup) from the response.
- **Cleanup**: the dead `get_address_from_lat_lon` Nominatim helper in `app/utils.py` (no callers) is removed together with its `requests` import.

Note: on the very first call after deploy, pg-nearest-city loads its bundled dataset into the database (one-time, ~30s+); subsequent calls are sub-millisecond.

## AI Tool Usage

- [ ] No AI tools were used

**If AI-assisted:**

- Tool(s) used: Claude Code (Anthropic)
- What was generated: initial implementation of the endpoint, frontend service swap, and the two route tests
- What you reviewed and changed: reviewed every change; verified the pg-nearest-city 2.0.0 API (Location fields, AsyncNearestCity accepting an existing psycopg connection) against the package source; ran the full backend test suite in the compose test stack and the frontend typecheck locally

## Screenshots

No UI change — `projectCountry` is populated exactly as before, from the new endpoint.

- Keeping the lookup in the frontend (status quo) means every project creation hits the shared OSM Nominatim server — the thing #554 wants to avoid.
- Computing city/country server-side only at project creation time wouldn't work here: the frontend needs the country *during* the create form (drone-altitude regulations lookup), so a small lookup endpoint fits better.

## Review Guide

- `pytest tests/test_projects_routes.py` — includes 2 new tests for the endpoint (`test_get_nearest_city`, `test_get_nearest_city_no_match`, geocoder faked to avoid dataset initialisation in CI).
- Full suite in the compose test stack: 341 passing locally (the two `test_batch_delete` jobs need `arq-worker` up).
- Manual: create a project and watch `/api/projects/nearest-city/` be called instead of `nominatim.openstreetmap.org`; first-ever call initialises the dataset.

## Checklist

- [x] PR is focused and small

### DIFF
--- a/src/backend/app/projects/project_routes.py
+++ b/src/backend/app/projects/project_routes.py
@@ -64,6 +64,7 @@ from fastapi import (
 from fastapi.responses import StreamingResponse
 from geojson_pydantic import FeatureCollection
 from loguru import logger as log
+from pg_nearest_city import AsyncNearestCity
 from psycopg import Connection
 from stream_zip import NO_COMPRESSION_64, stream_zip
 
@@ -109,6 +110,32 @@ async def read_project_centroids(
     return await project_logic.get_centroids(
         db,
     )
+
+
+@router.get("/nearest-city/", tags=["Projects"])
+async def get_nearest_city(
+    lon: float,
+    lat: float,
+    db: Annotated[Connection, Depends(database.get_db)],
+    user_data: Annotated[AuthUser, Depends(login_required)],
+):
+    """Reverse geocode a point to the nearest city and country.
+
+    Runs entirely offline against the PostGIS database via pg-nearest-city,
+    replacing the previous Nominatim-based lookup (reducing load on the
+    OSM-hosted service). The underlying dataset is loaded into the database
+    on the first call, which can take a little longer.
+    """
+    async with AsyncNearestCity(db) as geocoder:
+        location = await geocoder.query(lon, lat)
+
+    if location is None:
+        return {"city": None, "country": None}
+
+    return {
+        "city": location.city,
+        "country": location.country_name or location.country,
+    }
 
 
 @router.get("/{project_id}/download-boundaries", tags=["Projects"])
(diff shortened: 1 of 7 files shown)


Bloom79:

spwoodcock (changes requested):
We definitely want this - I support it!

But why do we need an endpoint returning nearest city?

Wouldn't be more logical to simply handle this on the backend when a project is made (as this is the only time the nearest city code needs to be called).

I can't imagine a scenario when the frontend needs to use this outside that context - let me know if you disagree! 🙏

PR #882, opened 2026-09-01
fix(s3): add_obj_to_bucket drops content_type, storing every upload as octet-stream

ppcvote:
`add_obj_to_bucket` declares and documents a `content_type`, but never sends it. The `put_object` call passes four positional arguments and then `**kwargs`, and because `content_type` is an explicit named parameter it binds there rather than staying in `kwargs`, so it cannot reach minio that way. Every BytesIO upload was therefore stored with the SDK's own `application/octet-stream` default.

Both callers are affected:

- `arq/tasks.py:1451` uploads the QField export as `application/zip`
- `projects/project_logic.py:146` passes the browser's `file.content_type` for user uploads

`metadata` was unaffected, since it does travel through `**kwargs`. The sibling `add_file_to_bucket` twelve lines above already forwards `content_type=` to `fput_object`.

The fix is one argument. The two tests record what `put_object` receives by binding the call through the real `Minio.put_object` signature, so what they assert is what the SDK would have put on the wire rather than what the wrapper thinks it sent. The second one pins that `metadata` still arrives, since the QField caller passes both.

Both fail on the current code and pass with the change. `ruff check` and `ruff format --check` are clean.

One note on how I ran them: this machine cannot install the full backend (GDAL and Scrapy do not build here), so I ran them with `--noconftest`, which is enough because these two touch nothing but `app.s3` and `minio`. They need no database, Redis or docker, so they should run normally in CI. If you would rather have this as an integration test against the real client, the equivalent assertion is `client.stat_object(bucket, key).content_type`, but I could not run that variant here and would not want to claim it verified.

### DIFF
--- a/src/backend/app/s3.py
+++ b/src/backend/app/s3.py
@@ -340,7 +340,12 @@ def add_obj_to_bucket(
     file_obj.seek(0)
 
     result = client.put_object(
-        bucket_name, s3_path, file_obj, file_obj.getbuffer().nbytes, **kwargs
+        bucket_name,
+        s3_path,
+        file_obj,
+        file_obj.getbuffer().nbytes,
+        content_type=content_type,
+        **kwargs,
     )
     log.debug(
         f"Created {result.object_name} object; etag: {result.etag}, "
--- a/src/backend/tests/test_s3_content_type.py
+++ b/src/backend/tests/test_s3_content_type.py
@@ -1,0 +1,65 @@
+"""add_obj_to_bucket declares a content_type, so it has to send one.
+
+The call it makes passes four positional arguments and then **kwargs. Because
+content_type is an explicit named parameter it is bound there rather than left
+in kwargs, so it never reached minio and every BytesIO upload was stored with
+the SDK's own application/octet-stream default.
+
+Arguments are recorded by binding them through the real Minio.put_object
+signature, so what is asserted is what the SDK would have put on the wire.
+"""
+
+import inspect
+from io import BytesIO
+
+from minio import Minio
+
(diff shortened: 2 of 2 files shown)


src/backend/app/s3.py
```diff
@@ -340,7 +340,12 @@ def add_obj_to_bucket(
     file_obj.seek(0)
 
     result = client.put_object(
-        bucket_name, s3_path, file_obj, file_obj.getbuffer().nbytes, **kwargs
```
spwoodcock, line 343: Would kwargs not encompass content_type if defined?

ppcvote: No, and that is the part that makes it easy to miss. `content_type` is an explicit named parameter in the signature, so Python binds the caller's value to it and it never reaches `**kwargs`:

```python
def demo(bucket, content_type="application/octet-stream", **kwargs):
    ...

demo("b", content_type="image/jpeg")
# content_type = 'image/jpeg'
# kwargs       = {}
```

Binding a real call through the current signature shows the same split: `content_type='image/jpeg'` sits in its own parameter while `kwargs={'metadata': {...}}`. So `metadata` did reach minio, which is why the QField upload's `Cache-Control` works today, and only `content_type` was dropped.

It would work through `**kwargs` if the parameter were removed from the signature, but then the documented default and the two callers passing it positionally would go with it. Forwarding it explicitly keeps both.

spwoodcock (approved):
Thanks 🙏 Can't merge unless you agree to the contribution policy, including AI guidelines

ppcvote:

PR #883, opened 2026-09-03
fix(waypoints): persist the default take off point instead of leaving NULL

### LINKED ISSUE #826: The takeoff point should always be saved, even if it's the default
- When we get a task summary for a project, it includes the takeoff point set by the user.
- However, in many cases the user just uses the default takeoff point (centroid of the task) and doesn't set it manually.
- In this case, the takeoff point is set to null.
- During processing, we need to know the location of the takeoff point, to (1) find the elevation from the global DEM (2) correct the (DJI) obsfucated z-value with a differential from the global DEM.
- It's not ideal having a bunch of `null` values for takeoff points. Even if the user only keeps the default, we should still write it to the db as a value!
- This way the task export will include a takeoff point for each flight.

This of course provides no guarantees the user actually used that flightplan, but we will come to that later.

Bloom79:

- [ ] 🍕 Feature
- [x] 🐛 Bug Fix
- [ ] 📝 Documentation
- [ ] 🧑‍💻 Refactor
- [x] ✅ Test
- [ ] 🤖 Build or CI

## Related Issue

Fixes #826

## Describe this PR

When a task's flight plan is generated without an explicit take off point, `get_task_flightplan` falls back to the task centroid, but only in memory: the task row kept `take_off_point = NULL`, so the task summary/export had no coordinate for flights that used the default (see #826).

This PR persists the centroid default with the existing `update_take_off_point_in_db` helper, the same way a user-supplied point is already persisted a few lines above. The precedence is unchanged: a supplied point wins and replaces whatever is stored, a stored point is reused, and only when neither exists is the centroid computed, and now saved.

As noted in the issue, this gives no guarantee the user actually flew that plan; it only guarantees the exported task always carries a take off point.

## AI Tool Usage

- [ ] No AI tools were used

**If AI-assisted:**

- Tool(s) used: Claude Code
- What was generated: the first draft of the `else` branch change in `waypoint_routes.py` and of the two regression tests.
- What you reviewed and changed: I traced the take off point flow myself (`update_take_off_point_in_db` / `get_take_off_point_from_db` in `task_logic.py`, the `waypoint_routes.py` branches, the `task_schemas.py` summary query) to confirm the fallback was the only place the default was left unsaved, checked the precedence between supplied, stored and default points, and ran the tests against the compose test stack.

## Screenshots

N/A (backend only).

- **Set the centroid at task creation (task splitting).** Would guarantee no NULL ever exists, but it writes a value before any flight plan is generated and would need a data migration for existing tasks. The issue explicitly ties the take off point to "each flight", so persisting at flight plan generation matches the intent with a much smaller change.
- **Compute the centroid in SQL (`ST_Centroid`) inside the update.** Works, but the route already has the shapely centroid in hand for `build_placemarks`; reusing that value guarantees the plan and the stored point are identical.

## Review Guide

- `src/backend/app/waypoints/waypoint_routes.py`: only the `else` branch of the take off point fallback changed (persist the centroid).
- `src/backend/tests/test_waypoint_routes.py`:
  - the task outline is now a module constant shared by a small `_setup_task` helper (the terrain follow helper reuses it, behaviour unchanged);
  - `test_default_take_off_point_is_persisted`: task starts with NULL, one call without a point, the stored point equals the shapely centroid and equals what `build_placemarks` received;
  - `test_supplied_take_off_point_replaces_persisted_default`: a persisted default is replaced by an explicit point.
- To run: `just test backend` (or `podman compose -f compose.test.yaml exec -T backend pytest tests/test_waypoint_routes.py -v`).

## How I tested it

I ran this branch on my local stack and tried it from the browser, as a drone operator would:

1. Opened a test project (582 tasks, all with `take_off_point = NULL`).
2. Locked task 1 and opened it. The map requested the flight plan on its own (`POST /api/waypoint/task/{id}/?download=false`, no `take_off_point` sent). I did not pick a take-off point: this is the #826 case.
3. Checked the DB: task 1 `take_off_point` is now set and equals `ST_Centroid(outline)`. Before it was NULL. Project count went from 0 to 1.

Note: the point is saved on the first request, so already on the map preview and even if that request then fails the DEM check (mine returned 409). The manual take-off point already works this way. If you'd rather save only after a successful download, it's a one-line change.

## Checklist

- [x] PR is focused and small

### DIFF
--- a/src/backend/app/waypoints/waypoint_routes.py
+++ b/src/backend/app/waypoints/waypoint_routes.py
@@ -101,9 +101,16 @@ async def get_task_flightplan(
         if take_off_point_from_db:
             take_off_point = take_off_point_from_db["coordinates"]
         else:
+            # No take off point supplied and none stored: default to the task
+            # centroid. Persist it so the task always carries a take off point
+            # (used to look up the DEM elevation during processing and included
+            # in task exports), instead of leaving NULL in the database.
             task_polygon = shape(task_geojson["features"][0]["geometry"])
             task_centroid = task_polygon.centroid
             take_off_point = [task_centroid.x, task_centroid.y]
+
+            geojson_point = {"type": "Point", "coordinates": take_off_point}
+            await update_take_off_point_in_db(db, task_id, geojson_point)
 
     # Flight params from project
     forward_overlap = project.front_overlap or 70
--- a/src/backend/tests/test_waypoint_routes.py
+++ b/src/backend/tests/test_waypoint_routes.py
@@ -2,24 +2,40 @@ import json
 import uuid
 
 import pytest
+from app.tasks.task_logic import get_take_off_point_from_db
 from app.waypoints import waypoint_routes
+from shapely.geometry import shape
+
+_TASK_OUTLINE = {
+    "type": "Polygon",
+    "coordinates": [
+        [
+            [-69.49779538720068, 18.629654277305633],
+            [-69.48497355306813, 18.616997544638636],
+            [-69.54053483430786, 18.608390428368665],
+            [-69.5410690773959, 18.614466085056165],
+            [-69.49779538720068, 18.629654277305633],
+        ]
+    ],
+}
+
+
+async def _setup_task(db, project_id: str) -> str:
+    """Insert a task with the shared test outline and no take off point."""
+    task_id = str(uuid.uuid4())
+    async with db.cursor() as cur:
+        await cur.execute(
+            """
+            INSERT INTO tasks (id, project_id, outline, project_task_index)
+            VALUES (%s, %s, ST_SetSRID(ST_GeomFromGeoJSON(%s), 4326), %s)
+            """,
+            (task_id, project_id, json.dumps(_TASK_OUTLINE), 1),
+        )
+    await db.commit()
+    return task_id
 
 
 async def _setup_terrain_follow_task(db, project_id: str) -> str:
-    task_id = str(uuid.uuid4())
-    outline = {
-        "type": "Polygon",
-        "coordinates": [
-            [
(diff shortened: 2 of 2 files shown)


spwoodcock:
Sorry but I can't keep reviewing these clearly AI generated PRs.

Thats how maintainers burn out 😅

https://responsibleai.guide

spwoodcock:
I know you probably mean well - thanks 🙏

But I have no reason to trust you have adequately reviewed what the AI has churned out yourself

Bloom79:
I understand,as openly reported in the comments I used AI for drafting and coding, I use it everyday at work, it's my "field". Anyway I review and test everything I deliver, in this case I'm running the app locally and testing all the PRs. For this one I locked a task, generated the flight plan without choosing a take-off point, and checked the DB row went from NULL to the task centroid.
I would really like to contribute to this project in a way that's useful rather than a burden. If you'd prefer to clear the queue, I'm happy to close #878 and #883. I really apologize if it created problems instead of helping. I fully understand your point anyway.

spwoodcock:
Thanks - really appreciate the disclosure 🙏
Otherwise I wouldn't even entertain checking the code.

Good to hear you tested it well 😄

Could you include that info in the PR next time?
I could generate a PR in 2 minutes, the same as you, but its the testing that takes real time!

---

And sorry to are taking the brunt. I maintain about 20 different repos and am inundated with AI PRs

Bloom79:
Thanks for the reply, I'm really happy to be of any support. I've added a "How I tested it" section to the PR description, with the steps I did in the browser and what I checked in the DB. I'll do the same in every PR from now on.

spwoodcock:
Thanks 😄 

I just tried the rotation on production & saw the ui to do this is broken. Did you also see this in testing?

